PR #12, opened 2021-10-01
Be more strict about missing fields

mstemm:
The biggest driver for this is to ensure that a user can write a
clause "and not ct.error exists" and have the clause evaluate to true
if there is no error. Previously, ct.error would return an empty
string, which still counted as "existing".

Instead of converting missing fields in the json object to "", be more
strict about missing fields:

- In Next(), if the eventTime or eventType are missing, skip the event
  by returning timeout.

- In getUser(), change the return value to (present, user) and set
  present to false if userIdentity is missing or if the code can't
  identify a user. In get_fields() for ct.user, set field_present to
  false if getUser()'s present is false. In event_to_string, use a
  blank string for the user if not present.

- In event_to_string, if eventSource, awsRegion, or eventName are
  missing, return an error string.

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area source-plugins

> /area extractor-plugins

> /area build

> /area documentation

Fixes #

```release-note
NONE
```

### DIFF
--- a/plugins/cloudtrail/cloudtrail.go
+++ b/plugins/cloudtrail/cloudtrail.go
@@ -729,10 +729,19 @@ func Next(plgState unsafe.Pointer, openState unsafe.Pointer) (*sdk.PluginEvent, 
 		return nil, sdk.SSPluginTimeout
 	}
 
+	// All cloudtrail events should have a time. If it's missing
+	// skip the event.
+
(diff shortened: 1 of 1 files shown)


leogr (approved):
/approve

PR #39, opened 2022-01-12
refactor(plugins): cleanup plugins, bump Go SDK version, and remove logging

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/area source-plugins

/area extractor-plugins

This PR cleans up the existing plugins as following:
- Renaming the `dummy_c` source name from `dummy` to `dummy_c`. Source names need to generally be unique, and although `dummy` and `dummy_c` are similar, the events they generate are not equal at all.
- Bump the Go SDK version
- Refactored and uniformed init configuration reading in `dummy` and `cloudtrail`, also aiming to avoid state variable replication
- Removing legacy debug logging, which caused a performance penalty in hot-paths even if the messages get discarded. They are also unused here and remained as legacy code. A SDK-native alternative solution, as described in https://github.com/falcosecurity/plugin-sdk-go/issues/24, would be better for this purposes.

```release-note
refactor(plugins): cleanup plugins, bump Go SDK version, and remove logging
```


leogr (approved):
/approve

PR #40, opened 2022-01-13
Refactor/new registry format

### LINKED ISSUE #34: improvements to the plugin registry
**Motivation**

I see a big potential in the plugin registry, but I think we might want to elaborate more on what it really is and how we should implement it. Currently, what we do is filling a markdown document with the publicly known plugins, which also help to track the list of assigned unique source IDs. I think we might want to further elaborate this concept, clarify some points, and reach a more sophisticated level.

**Feature**

This is not a real feature request, but instead I'm going to list some points and concerns that I would like to have a debate on:

1. **Clarify which plugins stay in the registry, and which not**. I think we need to formalize what a "known plugin" is. Are those the official ones, supported/maintained by the `falcosecurity` organization? Are these the ones that are distributed with the Falco binaries (if not, which are bundled in the distributions and which not)? Are these only open source, or do we also list proprietary plugins? Does this include only plugins in this repo, or also external ones.
2. **Is this the official source ID knowledge base?** Source plugins should not have overlapping IDs, and we need a way for developers to reserve those IDs. Is this the official place where those are collected? If so, how does the assignment process happen (where and how do we formally document it)? Does this need to happen for propertary plugins too, and how? Do we have similar needs for the other plugin types (e.g. extractor)?
3. **Clarify which information are collected for each plugin**. So far, for each plugin we collect `Name`, `Description`, `Contact`, `Event ID`, `Event Source`, `Extract Event Sources`. These are basically a portion of the info that plugins return to the plugin framework. Do we want more than this? What are the formal differences between the source and extractor plugin info?
4. **Switching to a machine-readable format**. I like the idea of tracking the plugins in a single location, but I don't think a MD document is the way to go. I would prefer having some machine-processable format (even a simple JSON would suffice) on which we could generate richer documents automatically, and that can be used in tools ad CI workflows.
5. **Publishing this on `falco-website`**. I think this should have a dedicated page on the official Falco website, so that users can fastly consult which plugins are available.

**Alternatives**

**Additional context**

andreabonanno:
**What type of PR is this?**

/kind feature

/area documentation

This PR addresses some points raised via issue #34, providing:

- single point of information in machine-readable format for the plugins relevant information, in the form of `registry/registry.yaml`
- script for checking ID and event sources conflicts in `registry.yaml` and usage of reserved sources names
- script to update this repo `README.md`, starting from `registry.yaml` 

Note: `registry.go` should be executed each time the registry is updated, to have `README.md` reflect the changes.

Fixes #34 

```release-note
NONE
```


README.md
```diff
@@ -8,5 +8,24 @@ This repository contains several reference [plugins](https://deploy-preview-493-
...
 
-When ready to release your plugin, make sure to register the plugin with the Falcosecurity organization by creating a PR to modify the [PLUGINS-REGISTRY.md](https://github.com/falcosecurity/plugins/blob/master/plugins/PLUGINS-REGISTRY.md) file with details on the new plugin. This ensures that a given ID is used by exactly one source plugin, and allows source plugin authors and extractor plugin authors to coordinate about event source formats.
+When ready to release your plugin, make sure to register the plugin with the Falcosecurity organization by creating a PR to modify the [registry.yaml](https://github.com/andreabonanno/plugins/blob/refactor/new-registry-format/registry/registry.yaml) file with details on the new plugin. This ensures that a given ID is used by exactly one source plugin, and allows source plugin authors and extractor plugin authors to coordinate about event source formats.
+
+# Plugin Registry
+
+<!-- REGISTRY -->
```
leogr, line 15: I believe it worth documenting this special comment with another comment here :)

build/registry/check.go
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	"fmt"
+	"os"
+	"regexp"
+)
+
+var (
+	forbiddenSourceIDs   = []uint{0}
+	forbiddenSourceNames = []string{"syscall", "internal", "plugins"}
```
leogr, line 27: It would be better to fetch that list from the `registry.yaml` file. 
Anyway, not a blocker for merging this PR.

registry.yaml
```diff
@@ -0,0 +1,55 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+plugins:
+  source:
+    - id: 1
+      source: k8s_audit
+      name: k8s_auditlogs
```
leogr, line 23:
```suggestion
      name: k8s_audit
      description: Reserved for a future back-port of Falco's k8s_audit event source as a plugin
```

leogr (commented):
Great!

I have left just a few comments, otherwise SGTM!

build/registry/types.go
```diff
@@ -0,0 +1,63 @@
...
+
+type Source struct {
+	ID          uint   `yaml:"id"`
+	Source      string `yaml:"source"`
+	Name        string `yaml:"name"`
+	Description string `yaml:"description"`
+	Authors     string `yaml:"authors"`
+	Contact     string `yaml:"contact"`
+	Repository  string `yaml:"repository"`
+	Reserved    bool    `yaml:"reserved"`
```
jasondellaluce, line 33: Tags seem to be not correctly aligned here. The same goes for lines 43 and 52/53.
Did the Go linter run in this file?

jasondellaluce (commented):
Left a comment, but other than that I think this is in a good shape and can be merged soon! Thanks @andreabonanno !

leogr (approved):
/approve

:clap:

PR #55, opened 2022-01-21
Fix off-by-one error in NextBatch

mstemm:
NextBatch is presented a long array of sdk.EventWriters and associates
events with them until the first error (which could include
eof/timeout).

If the last event results in an error, the loop would increment n,
then check the error, then return. n would be one bigger than
expected.

Fix this by checking for error + breaking inside the loop instead of
checking for error in the loop condition.

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area plugins

> /area registry

> /area build

> /area documentation

Fixes #

### DIFF
--- a/plugins/cloudtrail/cloudtrail.go
+++ b/plugins/cloudtrail/cloudtrail.go
@@ -211,8 +211,12 @@ func (o *openContext) NextBatch(pState sdk.PluginState, evts sdk.EventWriters) (
 	var n int
 	var err error
 	pCtx := pState.(*pluginContext)
-	for n = 0; err == nil && n < evts.Len(); n++ {
+	for n = 0; n < evts.Len(); n++ {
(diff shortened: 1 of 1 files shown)


FedeDP (approved):

FedeDP:
Uh, i didn't think i could really approve this given i am not a maintainer on this repo @leogr is that ok?

leogr (commented):
Hey @mstemm 

Good catch!

Could you also bump the plugin version to 0.2.2 (so that we can rerelease it)?

jasondellaluce (commented):
LGTM

Thanks for catching this @mstemm. Yeah I think once the version gets bumped we should be good to go!

leogr (approved):
/approve

PR #58, opened 2022-01-27
WIP: Return success on partial eof in NextBatch

mstemm:
If fetching a block of events stopped due to an eof, the old version
of NextBatch would return those events + an eof. This led to not
reading the last block of events at all.

Change this to return no error. Presumably, the next call to
NextBatch() would return EOF.

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area plugins

> /area registry

> /area build

> /area documentation

Fixes #

### DIFF
--- a/plugins/cloudtrail/cloudtrail.go
+++ b/plugins/cloudtrail/cloudtrail.go
@@ -218,6 +218,14 @@ func (o *openContext) NextBatch(pState sdk.PluginState, evts sdk.EventWriters) (
 			break
 		}
 	}
+
+	// If the loop above ended on EOF, but retrieved some events
+	// before EOF, return nil here. (The next call will return
+	// EOF).
+	if n > 0 && err == sdk.ErrEOF {
+		err = nil
+	}
+
 	return n, err
 }
 


mstemm:
I'm not *exactly* sure we need this change--should we assume the plugin framework will interpret a non-empty list of events + EOF to still act on any events that were returned?

That's why it's WIP.

jasondellaluce:
As you pointed out, I think this is not an issue. Actually, we always assumed this semantics was intended. We documented the behavior of NextBatch in the SDK Go as follows: https://pkg.go.dev/github.com/falcosecurity/plugin-sdk-go/pkg/sdk#NextBatcher

The idea is that `next_batch` can return a `# of events <= the batch size`. Once an EOF is returned, the framework consumes all the events from the current batch. What really matters is that, by contract, once the plugins return an EOF they needs to keep returning EOFs if `next_batch` is called again. Looking at the code, this also looks like the behavior implemented in [scap.c](https://github.com/falcosecurity/libs/blob/master/userspace/libscap/scap.c#L1822).

What do you think. Are we aligned on this?

mstemm:
Great, I'll close this then. Thanks for double-checking my logic.

PR #60, opened 2022-01-27
Update desc text for ct.error (not found value)

mstemm:
@ldegio pointed out that `ct.error` will return "field not present" if
the object does not have an `errorCode` property.

From the perspective of rules files, this ends up being the string
"&lt;NA&gt;", as that's how a NULL filtercheck value is represented as a
string.

This fixes the help text to reflect that.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

/kind documentation

> /kind failing-test

> /kind feature

/area plugins

> /area registry

> /area build

> /area documentation

Fixes #


leogr (approved):
/approve

:+1:

PR #64, opened 2022-02-02
fix(registry): remove uniqueness constraint on source name

jasondellaluce:

**What type of PR is this?**

/kind design

/area registry

/area documentation

This removes the uniqueness constraint that we used to put on the source name of plugins. I makes sense for plugins with different IDs to share the same source name **if they share the same event data format**. This can allow different plugin implementations of the same event source to co-exist in the ecosystem (and in the registry).


leogr (approved):
/approve

PR #68, opened 2022-02-08
docs: add a doc page about best practices for plugin development

jasondellaluce:
Co-authored-by: Leonardo Grasso <me@leonardograsso.com>

**What type of PR is this?**

/kind documentation

/area documentation

This introduces a `docs` directory meant to become the right place of pieces of documentation about this repository and plugins in general. As a first contribution, this adds a doc page for best/bad practices for plugin development.


docs/best-practices.md
```diff
@@ -0,0 +1,82 @@
...
+
+- Pass credentials or secrets in the plugin:
+    - Init configuration
+    - Open parameters
+- Create custom conventions (e.g. storing the secrets in a non-customizable path such as `~/.mysecrets`)
+- Embed credentials or secrets inside the plugin code
+
+### Do
+
+- Retrieve credentials or secrets through environment variables, or external files, and pass their name in the plugin:
```
leogr, line 30:
```suggestion
- Retrieve credentials or secrets through environment variables or external files, and pass their name in the plugin:
```

docs/best-practices.md
```diff
@@ -0,0 +1,82 @@
...
+    - Init configuration
+    - Open parameters, only if they are required to open the event source
+- Follow the secret management guidelines imposed by well-known packages such as [aws-sdk-go](https://github.com/aws/aws-sdk-go#configuring-credentials)
```
leogr, line 33:
```suggestion
- Follow the secret management guidelines mandated by well-known packages such as [aws-sdk-go](https://github.com/aws/aws-sdk-go#configuring-credentials)
```

docs/best-practices.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+## Managing the NextBatch Loop
+
+In source plugins, new events are created and returned in batches with the `plugin_next_batch()` function. This design decision reduces the overhead that may occur by integrating C with other programming languages such as Go. Although the SDKs cover most of the hard work, plugin developers must manage the batching system manually and there are some things to keep in mind in order to ensure optimal performance and correct behavior.
```
leogr, line 38:
```suggestion
In source plugins, new events are created and returned in batches with the `plugin_next_batch()` function. This design decision reduces the overhead that may occur by integrating C with other programming languages such as Go. Although the SDKs cover most of the hard work, plugin developers must manage the batching system manually and there are some things to keep in mind to ensure optimal performance and correct behavior.
```

docs/best-practices.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Don't
+
+- Allocate a new event batch at every invokation of `plugin_next_batch()`, because event generation is a very hot path
+- Block the execution for long times: be mindful that `plugin_next_batch()` is blocking for the framework
+- Return an EOF error and then return new events at the following invokation of `plugin_next_batch()`
+- Return a given number of events without filling their data in the event batch
```
leogr, line 45:
```suggestion
- Allocate a new event batch at every invocation of `plugin_next_batch()`, because event generation is a very hot path
- Block the execution for long times: be mindful that `plugin_next_batch()` is blocking for the framework
- Return an EOF error and then return new events at the following invocation of `plugin_next_batch()`
- Return a given number of events without filling their data in the event batch
```

docs/best-practices.md
```diff
@@ -0,0 +1,82 @@
...
+- Return Timeout (see [`sdk.ErrTimeout`](https://github.com/falcosecurity/plugin-sdk-go/blob/0b4b6dc215141116c53398f3232aac98e49cdb80/pkg/sdk/sdk.go#L30) in Go or [`SS_PLUGIN_TIMEOUT`](https://github.com/falcosecurity/libs/blob/033c4b9f28e58e20a5822bd8a7419beea098af91/userspace/libscap/plugin_info.h#L76) in C/C++) after a short time period passes without producing any event (no more than few milliseconds)
+- Return Timeout even with a non-zero number of events, if they are available
+- Return EOF (see [`sdk.ErrEOF`](https://github.com/falcosecurity/plugin-sdk-go/blob/0b4b6dc215141116c53398f3232aac98e49cdb80/pkg/sdk/sdk.go#L25) in Go or [`SS_PLUGIN_EOF`](https://github.com/falcosecurity/libs/blob/033c4b9f28e58e20a5822bd8a7419beea098af91/userspace/libscap/plugin_info.h#L77) in C/C++) when you are certain that no more events will be produced
+- Return EOF with a non-zero number of events, but ensuring that every following invokation of `plugin_next_batch()` returns EOF and zero events
```
leogr, line 54:
```suggestion
- Rely on the Go SDK and C++ SDK to ensure that all memory allocations are optimized
- Return a number of events smaller than the size of the batch, if fastly available (see the behavior described [in the Go SDK](https://pkg.go.dev/github.com/falcosecurity/plugin-sdk-go@v0.1.0/pkg/sdk#NextBatcher))
- Return Timeout (see [`sdk.ErrTimeout`](https://github.com/falcosecurity/plugin-sdk-go/blob/0b4b6dc215141116c53398f3232aac98e49cdb80/pkg/sdk/sdk.go#L30) in Go or [`SS_PLUGIN_TIMEOUT`](https://github.com/falcosecurity/libs/blob/033c4b9f28e58e20a5822bd8a7419beea098af91/userspace/libscap/plugin_info.h#L76) in C/C++) after a short time period passes without producing any event (no more than a few milliseconds)
- Return Timeout even with a non-zero number of events, if they are available
- Return EOF (see [`sdk.ErrEOF`](https://github.com/falcosecurity/plugin-sdk-go/blob/0b4b6dc215141116c53398f3232aac98e49cdb80/pkg/sdk/sdk.go#L25) in Go or [`SS_PLUGIN_EOF`](https://github.com/falcosecurity/libs/blob/033c4b9f28e58e20a5822bd8a7419beea098af91/userspace/libscap/plugin_info.h#L77) in C/C++) when you are certain that no more events will be produced
- Return EOF with a non-zero number of events, but ensuring that every following invocation of `plugin_next_batch()` returns EOF and zero events
```

docs/best-practices.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+## Fields Naming Conventions
+
+Plugins can extend Falco by supporting new fields to be extracted event data. At runtime, those fields co-exist with the other fields natively supported in libsinsp, such as the `evt.*` or the `container.*` classes. In this perspective, it's really important for plugin developers to follow some naming conventions for the newly introduced fields.
+
+### Don't
+
+- Introduce fields with the same name or the same class (e.g. `evt.*`, `container.*`, `ka.*`) as the ones already existing in libsinsp
```
leogr, line 63: This may contrast with https://falco.org/docs/plugins/#handling-duplicate-overlapping-fields-in-plugins-libraries-core

docs/best-practices.md
```diff
@@ -0,0 +1,82 @@
...
+- Use the `.` divisor character to give better meaning to your fields and divide them into categories (e.g. `ct.user.accountid` refers to the `user` category of AWS Cloudtrail events)
+
+
+## Live Events vs. Offline Events
+
+Once you get started developing a source plugin, you may come to the conclusion that there is absolutely no technical constraint on the events returned with `plugin_next_batch()`, since the plugin dictates both the data format of the events and how frequently they are produced. Some many degrees of freedom can be confusing and can make developer wonder to which extent the behavior of Falco should be changed.
+
+An important topic is the difference between producing events from a *live* or *offline* source. Those two terms are meant to distinguish events produced in real-time from the ones generated by parsing a static and reproducible source (such as a log file). In other words, live events are not reproducible because they are tied to real-time observations. To clear some of these doubts, be mindful of the following points:
```
leogr, line 78: May we want to give the readers a simpler definition of *live* events too?

I would say something like:

live events do not include events that happened before the plugin startup.

And give also some examples. For instance, assuming one has a log file:
- `tail -f` would output live events only
- `cat` would output all events (ie. offline mode)

leogr (commented):
Almost perfect! :star_struck: 

Just found some nits!

docs/best-practices.md
jasondellaluce: In my opinion it doesn't. The docs at falco.org are more tied to technical details, whereas this section is about best practices. In other words, this line does not imply that developers **can't** introduce overlapping filterchecks (which is technically feasible for different sources), but just state that it's better to not create ambiguity by introducing new fields with the same names as the ones of libhsinsp.

jasondellaluce:
Thanks @leogr! Added all of them.

leogr (approved):
/approve

PR #71, opened 2022-02-17
Cloudtrail: Handle floating point values.

geraldcombs:
Many of the events in a Cloudtrail scap here have floating point
bytesTransferred{In,Out} values, e.g.

"additionalEventData": {
  "SignatureVersion": "SigV4",
  "CipherSuite": "ECDHE-RSA-AES128-GCM-SHA256",
  "bytesTransferredIn": 0.0,
  "AuthenticationMethod": "AuthHeader",
  "x-amz-id-2": "Y/9E1+wcb4G90kzKPJ9K66wa+AIGssOLXHWQ7isdbcNc2OzUGBTYH4I7zjeZ3AR2zjn0oTuLZI4=",
  "bytesTransferredOut": 33.0
  }

This is arguably a bug with Cloudtrail, but here we are. Try parsing
them as Uint64 first, then fall back to Float64.

Signed-off-by: Gerald Combs <gerald@wireshark.org>

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area plugins

> /area registry

> /area build

> /area documentation

We were parsing S3 byte values using fastjson's GetUint64. Unfortunately, [that will return 0](https://github.com/valyala/fastjson/blob/6dae91c8e11a7fa6a257a550b75cba53ab81693e/fastfloat/parse.go#L82) if we try to feed it a floating point value, and Cloudtrail's byte values can be encoded as floats.

Fixes #

### DIFF
--- a/plugins/cloudtrail/cloudtrail.go
+++ b/plugins/cloudtrail/cloudtrail.go
@@ -54,7 +54,7 @@ const (
 	PluginName                      = "cloudtrail"
 	PluginDescription               = "reads cloudtrail JSON data saved to file in the directory specified in the settings"
 	PluginContact                   = "github.com/falcosecurity/plugins/"
-	PluginVersion                   = "0.2.4"
+	PluginVersion                   = "0.2.5"
 	PluginEventSource               = "aws_cloudtrail"
 )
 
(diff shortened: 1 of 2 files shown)


leogr (approved):
Good catch! Thank you :+1: 

I've also added a commit to bump the plugin version so that we can release it.

/approve

PR #72, opened 2022-02-17
wip: refactor(plugins/json): turn json plugin in a library and add unit tests

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind design

/area plugins

This PR proposes turning the json plugin in a library, so that it can be imported elsewhere to implement basic json field extraction functionalities. The rationale behind this proposal is that many plugins will need to extract fields from JSON data, and we can bundle all the related logic in a single place. Plus, unit tests have been added.

### DIFF
--- a/plugins/json/json.go
+++ b/plugins/json/json.go
@@ -21,17 +21,12 @@ limitations under the License.
 package main
 
 import (
-	"bytes"
-	"encoding/json"
 	"fmt"
-	"io"
-	"io/ioutil"
-	"strings"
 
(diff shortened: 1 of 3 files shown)


jasondellaluce:
Closing this in favor of https://github.com/falcosecurity/plugins/pull/82. I think fixing the support to JSON pointers and adding unit tests is the most crucial contribution here. Plus, while starting the development of the K8S Audit Log (https://github.com/falcosecurity/plugins/pull/79) plugin I realized that generalizing the JSON parsing tasks in a library package is not trivial. I propose to retry turning the `json` plugin in a shared package in the future.

PR #77, opened 2022-03-10
update(build/registry): adds URL plugin artifacts field in the registry

andreabonanno:
Signed-off-by: Andrea Bonanno <andrea@bonanno.cloud>
Co-Authored-By: Lorenzo Susini <susinilorenzo1@gmail.com>

**What type of PR is this?**

/kind feature

 /area registry

This PR adds the `Artifacts` field in the plugin registry, so we can differentiate bewteen sources and artifacts URLs.
As the plugin registry is the only source of truth for officially recognized plugins, this change could be useful to keep track where each artifact is published, an useful info for future consumers.

Fixes #

### DIFF
--- a/registry.yaml
+++ b/registry.yaml
@@ -36,7 +36,7 @@ plugins:
       description: Reads Cloudtrail JSON logs from files/S3 and injects as events
       authors: The Falco Authors
       contact: https://falco.org/community
-      url: https://github.com/falcosecurity/plugins/tree/master/plugins/cloudtrail
+      sources: https://github.com/falcosecurity/plugins/tree/master/plugins/cloudtrail
       license: Apache-2.0
     - id: 3
       source: dummy
@@ -44,7 +44,7 @@ plugins:
       description: Reference plugin used to document interface
       authors: The Falco Authors
       contact: https://falco.org/community
-      url: https://github.com/falcosecurity/plugins/tree/master/plugins/dummy
+      sources: https://github.com/falcosecurity/plugins/tree/master/plugins/dummy
       license: Apache-2.0
     - id: 4
       source: dummy_c
@@ -52,7 +52,7 @@ plugins:
       description: Like Dummy, but written in C++
       authors: The Falco Authors
       contact: https://falco.org/community
-      url: https://github.com/falcosecurity/plugins/tree/master/plugins/dummy_c
+      sources: https://github.com/falcosecurity/plugins/tree/master/plugins/dummy_c
       license: Apache-2.0
     - id: 5
       source: docker
@@ -60,7 +60,7 @@ plugins:
       description: Docker Events
       authors: Thomas Labarussias
       contact: https://github.org/Issif
-      url: https://github.com/Issif/docker-plugin
+      sources: https://github.com/Issif/docker-plugin
       license: Apache-2.0
     - id: 6
       source: seccompagent
@@ -68,7 +68,7 @@ plugins:
       description: Seccomp Agent Events
       authors: Alban Crequy
       contact: https://github.com/kinvolk/seccompagent
-      url: https://github.com/kinvolk/seccompagent
+      sources: https://github.com/kinvolk/seccompagent
       license: Apache-2.0
     - id: 999
       source: test
@@ -81,5 +81,5 @@ plugins:
       description: Extract values from any JSON payload
       authors: The Falco Authors
       contact: https://falco.org/community
-      url: https://github.com/falcosecurity/plugins/tree/master/plugins/json
+      sources: https://github.com/falcosecurity/plugins/tree/master/plugins/json
       license: Apache-2.0
--- a/build/registry/table.go
+++ b/build/registry/table.go
@@ -43,33 +43,35 @@ func (r *Registry) FormatMarkdownTable(contentType string) (string, error) {
 
 	switch contentType {
 	case sourcePluginsTableContentType:
-		ret.WriteString("| ID | Name | Event Source | Description | Info |\n")
-		ret.WriteString("| --- | --- | --- | --- | --- |\n")
+		ret.WriteString("| ID | Name | Event Source | Description | Info | Artifacts |\n")
+		ret.WriteString("| --- | --- | --- | --- | --- | --- |\n")
 		for _, s := range r.Plugins.Source {
-			line := fmt.Sprintf("| %d | %s | `%s` | %s | Authors: %s <br/> License: %s |\n",
+			line := fmt.Sprintf("| %d | %s | `%s` | %s | Authors: %s <br/> License: %s | %s |\n",
 				s.ID,
-				formatWithURL(s.Name, s.URL),
+				formatWithURL(s.Name, s.SourcesURL),
 				wrapNotAvailable(s.Source),
 				wrapNotAvailable(s.Description),
 				formatWithURL(s.Authors, s.Contact),
 				wrapNotAvailable(s.License),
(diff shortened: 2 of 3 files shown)


registry.yaml
```diff
@@ -36,39 +36,44 @@ plugins:
...
       license: Apache-2.0
     - id: 6
       source: seccompagent
       name: seccompagent
       description: Seccomp Agent Events
       authors: Alban Crequy
       contact: https://github.com/kinvolk/seccompagent
-      url: https://github.com/kinvolk/seccompagent
+      sources: https://github.com/kinvolk/seccompagent
+      url: https://download.falco.org/?prefix=plugins
```
leogr, line 76:
```suggestion
```
This plugin is not hosted at download.falco.org

registry.yaml
```diff
@@ -36,39 +36,44 @@ plugins:
...
       license: Apache-2.0
     - id: 5
       source: docker
       name: docker
       description: Docker Events
       authors: Thomas Labarussias
       contact: https://github.org/Issif
-      url: https://github.com/Issif/docker-plugin
+      sources: https://github.com/Issif/docker-plugin
+      url: https://download.falco.org/?prefix=plugins
```
leogr, line 67:
```suggestion
```
This plugin is not hosted at download.falco.org

leogr, line 67:
```suggestion
```
This plugin is not hosted at download.falco.org

leogr (changes requested):
Hey @andreabonanno 

I guess I understood your intent, but I have some concerns.  Are we sure the registry is the correct place to store artifacts-related information?

I thought we would need a separate index file (hosted somewhere, TBD) with all the metadata required to download the plugin files. Moreover, I see `url` as an informative link to the plugin's website (i.e., the project GitHub repository or the plugin website, if any). I don't think that `url` can be used as-is to download plugins. Am I wrong?

N.B. Not sure how you intend to consume that value, but you cannot use `https://download.falco.org/?prefix=plugins/stable/` (which is an AWS web UI for listing files) as a base URL for downloading files. If you want to use it as a base URL, you should use `https://download.falco.org/plugins/stable/` instead.

registry.yaml
andreabonanno: > Hey @andreabonanno
> 
> I guess I understood your intent, but I have some concerns. Are we sure the registry is the correct place to store artifacts-related information?
> 
> I thought we would need a separate index file (hosted somewhere, TBD) with all the metadata required to download the plugin files. Moreover, I see `url` as an informative link to the plugin's website (i.e., the project GitHub repository or the plugin website, if any). I don't think that `url` can be used as-is to download plugins. Am I wrong?
> 
> N.B. Not sure how you intend to consume that value, but you cannot use `https://download.falco.org/?prefix=plugins/stable/` (which is an AWS web UI for listing files) as a base URL for downloading files. If you want to use it as a base URL, you should use `https://download.falco.org/plugins/stable/` instead.

Hi and thanks for the feedback!
I do agree that the registry should not directly store artifact-related infos.
I envisioned the `url` field as a _pointer_ to separate index files, possibly hosted somewhere else, and those indexes in turn will be the source of truth for artifact information for that specific plugin. The registry would be the source of truth for those indexes, not for detailed artifacts infos. I can indeed refactor choosing a more expressive name than `url`, and changing the url used right now as a placeholder (as no indexes files present at the moment).

jasondellaluce:
Hi @andreabonanno, just leaving here some thoughs:
1. I think `sources` is ambiguous in this context. Here you seem to use it to indicate the source code location, however we also have a `source` field indicating the name of the event source for source plugins, and I think this can become confusing
2. `url` and `artifactURL` are used unevenly here. I think it would be much clearer to only pick one of the two
3. I think the use cases for artifact urls is limited in this context, and perhaps the registry might not even be the right place for that information. In order to have a frictionless distribution and availability of plugin artifacts, we would need at least the following:
        a. A base URL where the artifacts are hosted
        b. A way to define the version to download
        c. A way to define the platform (linux, windows, ...) to download
        d. A way to define the architecture (amd64, arm64, ...) to download
        
All those points kind of let me lean towards @leogr's idea of having a dedicated registry/index/hub just for artifacts.

andreabonanno:
> Hi @andreabonanno, just leaving here some thoughs:
> 
> 1. I think `sources` is ambiguous in this context. Here you seem to use it to indicate the source code location, however we also have a `source` field indicating the name of the event source for source plugins, and I think this can become confusing
> 2. `url` and `artifactURL` are used unevenly here. I think it would be much clearer to only pick one of the two
> 3. I think the use cases for artifact urls is limited in this context, and perhaps the registry might not even be the right place for that information. In order to have a frictionless distribution and availability of plugin artifacts, we would need at least the following:
>    a. A base URL where the artifacts are hosted
>    b. A way to define the version to download
>    c. A way to define the platform (linux, windows, ...) to download
>    d. A way to define the architecture (amd64, arm64, ...) to download
> 
> All those points kind of let me lean towards @leogr's idea of having a dedicated registry/index/hub just for artifacts.

@jasondellaluce I do agree on your points:
1-2) naming can definitely be improved to be less ambiguous
3) As I support the idea of @leogr, `artifactURL` is intended to be not just a pointer to a specific artifact, but one to the index/registry specific for that plugin. That is, the location where all the versions (platform/arch/...) of that plugin can be retieved. This location can be outside falcosecurity perimeter in the case of pluigns registered but hosted/maintained/published elsewhere. That being saied, I do agree that the name has to be improved.

leogr:
Hey @andreabonanno 

I'm still not convinced about this change. In particular:

1. I don't see any real benefit of having a dedicated field that points to the plugin codebase. I have assumed that `url` was the plugin website (which in most cases is also the source code repository website). That URL is mostly informative and intended for end-users. I don't see any advantage of changing that. I'd rather remove `contact` that's redundant IMO (@jasondellaluce shall we remove it from plugin API too? Or rename it to `url`?)
2. Ok for having a link to the artifact distribution system. I'd just introduce it later, once we have the actual distribution (index/repo/hub/whatever) in place. Also, I think the naming of `artifactURL` is too ambiguous. Perhaps `artifactsRepo` or `artifactsRepoURL` would be better. @andreabonanno @jasondellaluce any idea?

andreabonanno:
> Hey @andreabonanno
> 
> I'm still not convinced about this change. In particular:
> 
> 1. I don't see any real benefit of having a dedicated field that points to the plugin codebase. I have assumed that `url` was the plugin website (which in most cases is also the source code repository website). That URL is mostly informative and intended for end-users. I don't see any advantage of changing that. I'd rather remove `contact` that's redundant IMO (@jasondellaluce shall we remove it from plugin API too? Or rename it to `url`?)
> 2. Ok for having a link to the artifact distribution system. I'd just introduce it later, once we have the actual distribution (index/repo/hub/whatever) in place. Also, I think the naming of `artifactURL` is too ambiguous. Perhaps `artifactsRepo` or `artifactsRepoURL` would be better. @andreabonanno @jasondellaluce any idea?

@leogr thanks for the feedback.

1. I'm not opposed on having one single pointer for informative resources (either a plugin website or straight to the repo with the sources). In my opinion `contact` could be redundant because those info (emails mostly?) should be retrievable once one lands on the page pointed by `url`.
2. Yes maybe it's best to delay this field for later on, once it is backed up by a functional system for artifact distribution. I would go for `artifactsRepo` or `artifactsRepoURL` or something along those lines. Maybe stressing the plural ("artifactS") is important, as these location will host more than the last version of a plugin.

I will refactor this once we have a consensus!

PR #88, opened 2022-04-19
chore(build): allow publishing packages without README.md

jasondellaluce:

**What type of PR is this?**

/kind cleanup

/area build

This PR allows building and publishing plugin packages even if the README.md file is missing. So far, commands such as `make package/<plugin>` to fail if a readme is not found, which also blocks the publishing system of the CI. I still think putting readmes is valuable, but I don't see this as a publication requirement, especially for dev builds.


leogr (approved):
/approve

PR #119, opened 2022-06-07
fix(plugins/k8saudit): return non-nil error from NextBatch with closed channel

jasondellaluce:

**What type of PR is this?**

/kind bug

/area plugins

When the event source gets closed by the plugin due to lack of new events (this only happens when reading json events from file right now), the error returned by `NextBatch` was non-deterministically either `nil` or `sdk.ErrEOF`. In the current implementation of the plugin framework, returning empty event batches with `nil` error is interpreted as an unsafe corner case which causes a libscap-level failure. As a consequence, some Falco tests became flaky due to the program exit code being 1 or 0 with non-determinism. 

The root cause of this issue is the k8saudit plugin, which incorrectly handled a closing channel, thus returning a `nil` error instead of `sdk.ErrEOF`.

The `k8saudit` plugin version has been bumped to `v0.2.1` to include this patch.

If merged, the `k8saudit` plugin will need to be re-tagged and released with patch version bump to `v0.2.1`.

### DIFF
--- a/plugins/k8saudit/pkg/k8saudit/source.go
+++ b/plugins/k8saudit/pkg/k8saudit/source.go
@@ -211,7 +211,10 @@ func (k *Plugin) openEventSource(ctx context.Context, eventChan <-chan []byte, e
 				}
 			case <-ctx.Done():
 				return
-			case err := <-errorChan:
+			case err, ok := <-errorChan:
+				if !ok {
+					return
+				}
 				newErrorChan <- err
 			}
 		}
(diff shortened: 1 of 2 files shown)


plugins/k8saudit/pkg/k8saudit/source.go
```diff
@@ -211,7 +211,10 @@ func (k *Plugin) openEventSource(ctx context.Context, eventChan <-chan []byte, e
 				}
 			case <-ctx.Done():
 				return
-			case err := <-errorChan:
+			case err, ok := <-errorChan:
+				if !ok {
+					return
+				}
```
leogr, line 217: Although this works because those channels are private, I think just one channel should be responsible for shutting down the goroutine. Anyway, it's ok for now. I have left this comment just as a reminder for future improvements.

leogr (approved):

PR #120, opened 2022-06-09
refactor(build/registry): turn registry tool into an importable package

jasondellaluce:
**What type of PR is this?**

/kind design

/kind feature

/area registry

/area build

This PR refactors the `build/registry` tool to follow a package-oriented directory structured so that part of its code can be imported in external projects. The goal of the plugin registry is to be used in automatic tools in the future (such as https://github.com/falcosecurity/falcoctl/), and I think we should consider making this tool the single source of truth about the registry structure and contents.


leogr (approved):
Lovely

PR #121, opened 2022-06-10
wip: refactor(build/utils): add code for plugin loader and export package

jasondellaluce:
**What type of PR is this?**

/kind feature

/area build

This `build/version` tool, which is used to generate plugins packages in our CI, bundled some code logic that loaded plugin dynamic libraries at runtime. This PR generalizes that logic into an importable package that allows to load plugins and read their basic information. The `build/version` has been modified and simplified to use this package. This also adds unit tests to the newly-added package.

This is a first proposal towards the idea of having a community-supported plugin loader, which can be used for testing as an additional developer tool. At the same time, this can serve for loading plugins in automated tools. One first use case could be to auto-generate plugin READMEs by loading them and reading their information.


jasondellaluce:
Closing this in favor of https://github.com/falcosecurity/plugin-sdk-go/pull/56

PR #131, opened 2022-07-13
add shared libs for AWS Session + Cloudwatch

Issif:
Signed-off-by: Issif <issif+github@gadz.org>

**What type of PR is this?**

> /kind bug

> /kind cleanup

/kind design

> /kind documentation

> /kind failing-test

 /kind feature

/area plugins

> /area registry

> /area build

> /area documentation

This PR introduces 2 libs for the Go SDK, they are useful to:
* create a session to AWS's API
* open an Instance for Cloudwatch logs

With this libs, it's easy to open an instance to retrieve logs from AWS

Fixes #


shared/go/aws/cloudwatchlogs/cloudwatch.go
```diff
@@ -0,0 +1,84 @@
...
+
+// CreateClient returns a Client for CloudwatchLogs API
+func CreateClient(cfgs *aws.Config) *Client {
+	return &Client{
+		CloudWatchLogs: cloudwatchlogs.New(session.CreateSession(), cfgs),
+	}
+}
+
+// CreateFilter returns a filter for the query to CloudwatchLogs API
+func CreateFilter(startTime, endTime int64, filterPattern, logGroupName, logStreamNamePrefix, logStreamNames string) *cloudwatchlogs.FilterLogEventsInput {
```
jasondellaluce, line 42: startTime and endTime seem to not be used here. Plus, I would make them of type `time.Time`. More broadly, I don't see much value in having this function, which just passes-in its parameters to `cloudwatchlogs.FilterLogEventsInput`.

shared/go/aws/cloudwatchlogs/cloudwatch.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+}
+
+// Open returns an instance with the functionn called to retrieve logs
+func (client *Client) Open(filters *cloudwatchlogs.FilterLogEventsInput, refresh int64) (source.Instance, error) {
+	eventC := make(chan source.PushEvent)
+
+	go func() {
+		defer close(eventC)
+		for {
```
jasondellaluce, line 58: What's the termination condition for this event loop?

shared/go/aws/cloudwatchlogs/cloudwatch.go
```diff
@@ -0,0 +1,84 @@
...
+			filters.SetNextToken("")
+			client.getLogEvents(filters, eventC)
+			time.Sleep(time.Duration(refresh) * time.Second)
+		}
+	}()
+	return source.NewPushInstance(eventC)
```
jasondellaluce, line 64: Similarly to the comment above, how does the CW client get closed? It would be good to have a "closing" callback to pass to the instance with the `source.WithInstanceClose` option.

shared/go/aws/cloudwatchlogs/cloudwatch.go
```diff
@@ -0,0 +1,84 @@
...
+		StartTime:           time.Now().Add(-10 * time.Second).UnixMilli(),
+		FilterPattern:       aws.String(filterPattern),
+		LogGroupName:        aws.String(logGroupName),
+		LogStreamNamePrefix: aws.String(logStreamNamePrefix),
+		LogStreamNames:      aws.String(logStreamNames),
+	}
+}
+
+// Open returns an instance with the functionn called to retrieve logs
+func (client *Client) Open(filters *cloudwatchlogs.FilterLogEventsInput, refresh int64) (source.Instance, error) {
```
jasondellaluce, line 53: A downside of creating a `source.Instance` internally is that developers have _no control_ over the data written on each event, which would always be the raw content of each CW log entry. For example, this can be limiting for cases like a future AKS integration, in which we may want to pre-process the log entries before writing them in the `k8saudit` event source.

What about passing an optional `func([]byte) []byte` transformation function? Alternatively, this could return a read-only `source.PushEvent` chan, so that users can choose to preprocess each event before pushing it to the event source.

shared/go/aws/cloudwatchlogs/cloudwatch.go
```diff
@@ -0,0 +1,84 @@
...
+	}()
+	return source.NewPushInstance(eventC)
+}
+
+func (client *Client) getLogEvents(filters *cloudwatchlogs.FilterLogEventsInput, eventC chan source.PushEvent) error {
+	logs, err := client.CloudWatchLogs.FilterLogEvents(filters)
+	if err != nil {
+		return sdk.ErrEOF
+	}
+	if len(logs.Events) == 0 {
```
jasondellaluce, line 72: Note that EOF will make the event source terminate, and eventually will make consumers like Falco exit. What if we want the poll for new log entries and wait for new ones to be written? I don't think this should be the termination condition (at least, this should be something configurable from the outside). In general, I think the right error to return here is `sdk.ErrTimeout`.

shared/go/aws/cloudwatchlogs/cloudwatch.go
```diff
@@ -0,0 +1,84 @@
...
+
+// Open returns an instance with the functionn called to retrieve logs
+func (client *Client) Open(filters *cloudwatchlogs.FilterLogEventsInput, refresh int64) (source.Instance, error) {
+	eventC := make(chan source.PushEvent)
+
+	go func() {
+		defer close(eventC)
+		for {
+			filters.SetNextToken("")
+			client.getLogEvents(filters, eventC)
```
jasondellaluce, line 60:
```suggestion
if err := client.getLogEvents(filters, eventC); err != nil && err != sdk.ErrTimeout {
    eventC <- source.PushEvent{Err: err}
    return
}
```
Or something like this. Errors are lost here

shared/go/aws/cloudwatchlogs/cloudwatch.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	if len(logs.Events) == 0 {
+		return sdk.EOF
+	}
+	for i := range logs.Events {
+		eventC <- source.PushEvent{Data: i}
+	}
+	if logs.NextToken != nil {
+		filters.SetNextToken(*logs.NextToken)
+		client.getLogEvents(filters, eventC)
```
jasondellaluce, line 80: Same here, errors are not catched. Plus, be careful with recursion! A loop would do the job here.

jasondellaluce (commented):
Cool! Some initial thoughts.

shared/go/aws/cloudwatchlogs/cloudwatch.go
Issif: I think the easiest way is to return a `chan []byte`, wdyt?

jasondellaluce: Yeah I agree. `source.PushEvent` can contain either a `[]byte` or a `error`, so it can be handier to propagate errors in the channel.

jasondellaluce (commented):
Looks good now! Just few minor comments/question.

I think a big question here is _how useful_ a simple wrapper around the AWS SDK might be a this point. However, I see the big picture and expanding this to many others pre-built definitions can be something quite useful for many. Looking forward to hear some feedback from others too!

shared/go/aws/cloudwatchlogs/cloudwatch.go
```diff
@@ -0,0 +1,79 @@
...
+			}
+			for _, i := range logs.Events {
+				eventC <- i
+			}
+			if logs.NextToken != nil {
+				filters.SetNextToken(*logs.NextToken)
+				continue
+			}
+			filters.SetNextToken("")
+			filters.StartTime(*logs[len(logs)-1].Timestamp + 1)
```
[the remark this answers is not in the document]
Issif: It's not supposed to happen, it would mean AWS has a big issue :sweat_smile:

shared/go/aws/cloudwatchlogs/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/falcosecurity/plugins/shared/go/aws/cloudwatchlogs
+
+go 1.18
```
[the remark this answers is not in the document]
Issif: it's not, I can lower it

shared/go/aws/cloudwatchlogs/cloudwatch.go
```diff
@@ -0,0 +1,79 @@
...
+func (client *Client) Open(context context.Context, filterPattern, logGroupName, logStreamNamePrefix, logStreamNames string, shift, refresh int64) (chan *cloudwatchlogs.FilteredLogEvent, chan error) {
+	filters := &cloudwatchlogs.FilterLogEventsInput{
+		StartTime:           time.Now().Add(-1 * time.Duration(shift) * time.Second).UnixMilli(),
+		FilterPattern:       aws.String(filterPattern),
+		LogGroupName:        aws.String(logGroupName),
+		LogStreamNamePrefix: aws.String(logStreamNamePrefix),
+		LogStreamNames:      aws.String(logStreamNames),
+	}
+
+	eventC := make(chan []*cloudwatchlogs.FilteredLogEvent, 200)
```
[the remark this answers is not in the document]
Issif: I will change that for a parameter

shared/go/aws/cloudwatchlogs/cloudwatch.go
```diff
@@ -0,0 +1,79 @@
...
+
+// CreateClient returns a Client for CloudwatchLogs API
+func CreateClient(cfgs *aws.Config) *Client {
+	return &Client{
+		CloudWatchLogs: cloudwatchlogs.New(session.CreateSession(), cfgs),
+	}
+}
+
+// Open returns an instance with the functionn called to retrieve logs
+func (client *Client) Open(context context.Context, filterPattern, logGroupName, logStreamNamePrefix, logStreamNames string, shift, refresh int64) (chan *cloudwatchlogs.FilteredLogEvent, chan error) {
```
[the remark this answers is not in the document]
Issif: I will create a struct for filters I think

shared/go/aws/cloudwatchlogs/cloudwatch.go
Issif: if the context is Cancelled, `FilterLogEventsWithContext` should returns an error and break the loop

shared/go/aws/cloudwatchlogs/cloudwatch.go
```diff
@@ -0,0 +1,125 @@
...
+// Client represents a client for CloudwatchLogs API
+type Client struct {
+	*cloudwatchlogs.CloudWatchLogs
+}
+
+// Options represents options for calls to CloudwatchLogs API
+type Options struct {
+	Shift      int64
+	Refresh    int64
+	BufferSize int64
```
jasondellaluce, line 51:
```suggestion
	BufferSize uint64
```

shared/go/aws/cloudwatchlogs/cloudwatch.go
```diff
@@ -0,0 +1,125 @@
...
+func (options *Options) setDefault() {
+	if options.Refresh <= 1 {
+		options.Refresh = defaultRefresh
+	}
+	if options.Shift < 0 {
+		options.Refresh = defaultShift
+	}
+	if options.BufferSize <= 0 {
+		options.Refresh = defaultBufferSize
+	}
```
jasondellaluce, line 81:
```suggestion
       if options.Refresh == 0 {
		options.Refresh = defaultRefresh
	}
	if options.Shift == 0 {
		options.Shift = defaultShift
	}
	if options.BufferSize == 0 {
		options.BufferSize = defaultBufferSize
	}
```
Note that you're just assigning options.Refresh here. Also, changing the option types as suggested above allows you to just check for their 0 values

shared/go/aws/cloudwatchlogs/cloudwatch.go
```diff
@@ -0,0 +1,125 @@
...
+
+// Client represents a client for CloudwatchLogs API
+type Client struct {
+	*cloudwatchlogs.CloudWatchLogs
+}
+
+// Options represents options for calls to CloudwatchLogs API
+type Options struct {
+	Shift      int64
+	Refresh    int64
```
jasondellaluce, line 50: nit: I think `PollingInterval` might be a more clear name than `Refresh`.

shared/go/aws/cloudwatchlogs/cloudwatch.go
```diff
@@ -0,0 +1,125 @@
...
+	"context"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/cloudwatchlogs"
+	"github.com/falcosecurity/plugins/shared/go/aws/session"
+)
+
+const (
+	defaultShift      int64 = 1   // controls the time shift in past for the first call to CloudwatchLogs API
```
jasondellaluce, line 29: I think it would make sense to have these public

shared/go/aws/cloudwatchlogs/cloudwatch.go
Issif: smart

shared/go/aws/cloudwatchlogs/cloudwatch.go
Issif: right

shared/go/aws/cloudwatchlogs/cloudwatch.go
```diff
@@ -0,0 +1,125 @@
...
+
+// Client represents a client for CloudwatchLogs API
+type Client struct {
+	*cloudwatchlogs.CloudWatchLogs
+}
+
+// Options represents options for calls to CloudwatchLogs API
+type Options struct {
+	Shift           time.Duration
+	PollingInterval uint64
```
jasondellaluce, line 50:
```suggestion
	PollingInterval time.Duration
```

PollingInterval might be a time.Duration too, so that users don't need to have knowledge about the measure unit.

Issif: :+1:

leogr (commented):
SGTM, but I feel like the [session package](https://github.com/falcosecurity/plugins/pull/131/files#diff-3fb797ff4c2510e524e6974dfd3194bebccb2ee06f5e14a376762260ea390010) is not needed nor useful.

Is there any compelling reason to keep it?

Issif:
> SGTM, but I feel like the [session package](https://github.com/falcosecurity/plugins/pull/131/files#diff-3fb797ff4c2510e524e6974dfd3194bebccb2ee06f5e14a376762260ea390010) is not needed nor useful.
> 
> Is there any compelling reason to keep it?

In my mind, it was useful to have a module for sessions, as this one could be used for authentication for different services of AWS. I'm using it for CloudwatchLogs, but it could be used to subscribe to SQS for example.

jasondellaluce (approved):
/approve

PR #132, opened 2022-07-16
fix(plugins/github): correctly parse patch diffs 

jasondellaluce:
**What type of PR is this?**

/kind bug

/area plugins

The patch diff parsing function was fragile and rejected cases like `"@@ -0,0 +1 @@\n+sample line\n"`. This has been fixed by using regular expressions.


leogr (approved):
:+1:

PR #134, opened 2022-07-26
EKS k8s audit logs plugin

Issif:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area plugins

> /area registry

> /area build

> /area documentation

This PR adds a new plugin for k8s audit logs for EKS from Cloudwatch, it is based on the `k8saudit` plugin for field extraction and on shared libs for AWS services, see #131 

Example:
```
root@falco:/etc/falco/plugins/k8sauditeks# falco -c falco.yaml -r rules/k8s_audit_rules.yaml 
15:12:43.826983000: Warning K8s Operation performed by user not in allowed list of users (user=eks:addon-manager target=coredns/serviceaccounts verb=patch uri=/api/v1/namespaces/kube-system/serviceaccounts/coredns?fieldManager=eks&force=true&timeout=10s resp=200)
15:12:43.829125000: Informational K8s Serviceaccount Created (user=system:kube-controller-manager serviceaccount=resourcequota-controller ns=kube-system resp=201 decision=allow reason=RBAC: allowed by ClusterRoleBinding "system:kube-controller-manager" of ClusterRole "system:kube-controller-manager" to User "system:kube-controller-manager")
```

This plugin depends of #131 to test it, you need to add `replace` in `go.mod`, and as I forked my branch for the shared libs, this current PR includes the changes for it too. After the merge of #131 it will be easier to review

Fixes #


plugins/k8sauditeks/Makefile
```diff
@@ -0,0 +1,46 @@
...
+else
+    GODEBUGFLAGS= GODEBUG=cgocheck=0
+endif
+
+all: $(OUTPUT)
+
+clean:
+	@rm -f *.so *.h
+
+test-files:
```
jasondellaluce, line 36: This is a copy-paste from k8saudit I guess! I think we should not include this for now to reduce entropy. This was put there just to provide a barebones benchmark attempt, but we need a better testing strategy in the general case.

plugins/k8sauditeks/Makefile
```diff
@@ -0,0 +1,46 @@
...
+
+SHELL=/bin/bash -o pipefail
+GO ?= go
+CURL ?= curl
+
+FALCOSECURITY_FALCO_REPO ?= falcosecurity/falco
+FALCOSECURITY_FALCO_REVISION ?= 65435d4418638176f68df9017df36cd5d94ba083
+
+TEST_FILES_DIR := test_files
+NAME := k8sauditeks
```
jasondellaluce, line 22: I believe a better name for this pluginnwould be `eks-k8saudit`. It's more readable and it sorts of hints that this is a "flavor" of k8saudit. It's also in line with the current sample docs where this was first envisioned and mentioned: https://falco.org/docs/plugins/#plugin-event-sources-and-interoperability

plugins/k8sauditeks/README.md
```diff
@@ -0,0 +1,79 @@
...
+plugins:
+  - name: k8sauditeks
+    library_path: libk8sauditeks.so
+    init_config:
+      region: "us-east-1"
+      profile: "default"
+      cluster_name: "my-cluster"
+      shift: 10
+      polling_interval: 10
+      use_async: false
```
jasondellaluce, line 40: Let's put this as true by default. It's a huge performance gain. See: https://github.com/falcosecurity/plugin-sdk-go/issues/62

plugins/k8sauditeks/README.md
```diff
@@ -0,0 +1,79 @@
...
+### Configuration
+
+Here's an example of configuration of `falco.yaml`:
+
+```yaml
+plugins:
+  - name: k8sauditeks
+    library_path: libk8sauditeks.so
+    init_config:
+      region: "us-east-1"
```
jasondellaluce, line 35: I see `region`, `profile`, and `cluster_name` as good candidates for open parameters instead than init configuration. Do you need these elsewhere out of the generation of events? If not, these are good open parameters for the event source.

Since that is usually a string, you should propose some URI-like encoding form.

plugins/k8sauditeks/README.md
```diff
@@ -0,0 +1,79 @@
...
+ * `profile`: The AWS Profile to use
+ * `cluster_name`: The name of your EKS Cluster
+ * `use_async`: If true then async extraction optimization is enabled (Default: true)
+ * `polling_interval`: Polling Interval in seconds (default: 5s)
+ * `shift`: Time shift in past in seconds (default: 1s)
+ * `buffer_size`: Buffer Size (default: 200)
+
+### Rules
+
+The `k8sauditeks` plugin ships with a default set of ruleset (see `rules/` directory). These rules are same than those for `k8saudit`, only the required plugin changes.
```
jasondellaluce, line 61: Replicating the whole ruleset is a bad choice here. This plugin should not ship any ruleset, and be compatible with the `k8saudit` one. I see why you couldn't do it, and I opened an issue for this problem specifically: https://github.com/falcosecurity/falco/issues/2150

So I think our choice here would be to:
- Block this PR until https://github.com/falcosecurity/falco/issues/2150 is merged
- Remove the `k8saudit` plugin dependency on the ruleset and keep only `json`, so that this plugin gets "implicitly" accepted too

plugins/k8sauditeks/falco.yaml
```diff
@@ -0,0 +1,14 @@
+plugins:
+  - name: k8sauditeks
+    library_path: /etc/falco/plugins/k8sauditeks/libk8sauditeks.so
+    init_config: '{"region": "us-east-1", "profile": "draios-dev-developer", "cluster_name": "RishZ", "polling_interval": 2}'
```
jasondellaluce, line 4: Leftovers? I see no point in having a sample falco.yaml here anyway since we have all this info in the README

plugins/k8sauditeks/go.mod
```diff
@@ -0,0 +1,12 @@
+module github.com/falcosecurity/plugins/plugins/k8saudit-eks
```
jasondellaluce, line 1: This is not consitent with the name of the directory. Consider my naming suggestion above here too.

plugins/k8sauditeks/pkg/k8sauditeks/k8sauditeks.go
```diff
@@ -0,0 +1,137 @@
...
+
+const pluginName = "k8sauditeks"
+
+type Plugin struct {
+	k8saudit.Plugin
+	Config PluginConfig
+}
+
+type PluginConfig struct {
+	Region          string `json:"region"           jsonschema:"description=The Region of your EKS cluster (default: us-east-1), it can be replaced by env var AWS_REGION"`
```
jasondellaluce, line 43: Same as before: Region, Profile, and Cluster name are good open parameters candidates.

plugins/k8sauditeks/pkg/k8sauditeks/k8sauditeks.go
```diff
@@ -0,0 +1,137 @@
...
+		k.Profile = p
+	}
+	if r := os.Getenv("AWS_DEFAULT_REGION"); r != "" {
+		k.Region = r
+	}
+	if r := os.Getenv("AWS_REGION"); r != "" {
+		k.Region = r
+	}
+	if k.Region == "" {
+		k.Region = "us-east-1"
```
jasondellaluce, line 78: I think region should not be optional here. Another point in favor of making it an open param.

plugins/k8sauditeks/pkg/k8sauditeks/k8sauditeks.go
```diff
@@ -0,0 +1,137 @@
...
+	if schema, err := reflector.Reflect(&PluginConfig{}).MarshalJSON(); err == nil {
+		return &sdk.SchemaInfo{
+			Schema: string(schema),
+		}
+	}
+	return nil
+}
+
+func (k *Plugin) Open(params string) (source.Instance, error) {
+	filter := cloudwatchlogs.CreateFilter("", "/aws/eks/"+k.Config.ClusterName+"/cluster", "kube-apiserver-audit", nil)
```
jasondellaluce, line 114: Maybe `kube-apiserver-audit` should be configurable too.

plugins/k8sauditeks/rules/k8s_audit_rules.yaml
```diff
@@ -0,0 +1,743 @@
+#
```
jasondellaluce, line 1: This file is a replica of the k8saudit one and I'm against having two copies, see comment above.

plugins/k8sauditeks/pkg/k8sauditeks/k8sauditeks.go
```diff
@@ -0,0 +1,137 @@
...
+		time.Duration(k.Config.PollingInterval*uint64(time.Second)),
+		k.Config.BufferSize,
+	)
+	eventsC, errC := client.Open(ctx, filter, options)
+	pushEventC := make(chan source.PushEvent)
+	go func() {
+		for {
+			select {
+			case i := <-eventsC:
+				pushEventC <- source.PushEvent{Data: []byte(*i.Message)}
```
jasondellaluce, line 128: You should use the `k8saudit.ParseAuditEventsPayload` method here before pushing to the channel. That would allow for content checking and timestamp setting. See how the `parseAuditEventsAndPush` private method is implemented (and how it logs errors).

jasondellaluce (commented):
Great work! I left a first round of comments

plugins/k8sauditeks/Makefile
Issif: I'll clean that.

plugins/k8sauditeks/Makefile
Issif: I prefer `k8saudit-eks`, it allows to have all `k8saudit-X` plugins listed side by side in the repository, wdyt?

plugins/k8sauditeks/README.md
Issif: ok

plugins/k8sauditeks/README.md
Issif: I'll switch them to `open_parameters`. 
What do you mean by "URI-like encoding form"?

plugins/k8sauditeks/README.md
Issif: I'm off for next 2 weeks, we can wait til my return

plugins/k8sauditeks/falco.yaml
Issif: :sweat_smile:

plugins/k8sauditeks/pkg/k8sauditeks/k8sauditeks.go
Issif: Ok

plugins/k8sauditeks/pkg/k8sauditeks/k8sauditeks.go
Issif: It's a string enforced by AWS for the logs.

plugins/k8sauditeks/pkg/k8sauditeks/k8sauditeks.go
Issif: :+1:

plugins/k8sauditeks/Makefile
jasondellaluce: Yeah that works for me!

plugins/k8sauditeks/pkg/k8sauditeks/k8sauditeks.go
Issif: For the region, I don't think we should make it required, I don't think either we should set a default value. Most users will run Falco on EKS, and by design, the SDK will access to the correct endpoints of AWS API. The region is only used for that. wydt?

plugins/k8saudit-eks/pkg/k8sauditeks/k8sauditeks.go
```diff
@@ -0,0 +1,185 @@
...
+const pluginName = "k8saudit-eks"
+
+type Plugin struct {
+	k8saudit.Plugin
+	Logger *log.Logger
+	Config PluginConfig
+}
+
+type PluginParams struct {
+	ClusterName string `json:"cluster_name"     jsonschema:"description=The name of your EKS Cluster,required"`
```
geraldcombs, line 45: Can you add a title for each of these as well?

plugins/k8saudit-eks/pkg/k8sauditeks/k8sauditeks.go
```diff
@@ -0,0 +1,185 @@
...
+	Profile     string `json:"profile"          jsonschema:"description=The Profile to use to create the session, env var AWS_PROFILE if present"`
+	Region      string `json:"region"           jsonschema:"description=The Region of your EKS cluster, env var AWS_REGION is used if present"`
+}
+
+type PluginConfig struct {
+	BufferSize      uint64 `json:"buffer_size"      jsonschema:"description=Buffer Size (default: 200)"`
```
geraldcombs, line 51: Can you add a title for each of these along with a "default=" annotation?

plugins/k8sauditeks/README.md
Issif: I replaced the ruleset by a single dummy rule to allow people to test the plugin, we could change that in future, wdyt?

jasondellaluce: I think the way to go for now would be to remove the `k8s_audit` requirement from `required_plugin_versions` of the actual k8saudit ruleset, so that users can test this against the real k8s audit rules and we don't have an unused dummy ruleset here.

Issif: I agree, but what can we do until that? I don't want to block this PR too long

plugins/k8saudit-eks/Makefile
```diff
@@ -0,0 +1,36 @@
+#
```
jasondellaluce, line 1: The makefile of other plugins have been changed recently to support automatic readme generation. See https://github.com/falcosecurity/plugins/pull/138. I would be great if we could have those changes here as well.

For the future: maybe a makefile template (at least for Go plugins) might be the way to go.

plugins/k8saudit-eks/pkg/k8sauditeks/k8sauditeks.go
```diff
@@ -0,0 +1,185 @@
...
+		time.Duration(p.Config.PollingInterval*uint64(time.Second)),
+		p.Config.BufferSize,
+	)
+	eventsC, errC := client.Open(ctx, filter, options)
+	pushEventC := make(chan source.PushEvent)
+	go func() {
+		for {
+			select {
+			case i := <-eventsC:
+				// fmt.Println(*i.Message)
```
jasondellaluce, line 160: Leftover?

plugins/k8saudit-eks/pkg/k8sauditeks/k8sauditeks.go
```diff
@@ -0,0 +1,185 @@
...
+					p.Logger.Println(err)
+					continue
+				}
+				for _, j := range values {
+					if j.Err != nil {
+						p.Logger.Println(j.Err)
+						continue
+					}
+					pushEventC <- *j
+					// pushEventC <- *j
```
jasondellaluce, line 172: Leftover?

plugins/k8saudit-eks/pkg/k8sauditeks/k8sauditeks.go
Issif: :sweat_smile:

plugins/k8saudit-eks/Makefile
Issif: :+1:

Issif: It's updated

plugins/k8saudit-eks/pkg/k8sauditeks/k8sauditeks.go
```diff
@@ -0,0 +1,163 @@
...
+		EventSource: "k8s_audit",
+	}
+}
+
+// Reset sets the configuration to its default values
+func (p *PluginConfig) Reset() {
+	if i := os.Getenv("AWS_DEFAULT_PROFILE"); i != "" {
+		p.Profile = i
+	}
+	if i := os.Getenv("AWS_PROFILE"); i != "" {
```
Kaizhe, line 70: if `AWS_PROFILE` is not presented, should we look up for other environment variable as well?

plugins/k8saudit-eks/pkg/k8sauditeks/k8sauditeks.go
```diff
@@ -0,0 +1,163 @@
...
+	return []sdk.OpenParam{
+		{Value: "default", Desc: "Cluster Name"},
+	}, nil
+}
+
+func (p *Plugin) Open(clustername string) (source.Instance, error) {
+	if clustername == "" {
+		return nil, fmt.Errorf("cluster name can't be empty")
+	}
+	filter := cloudwatchlogs.CreateFilter("", "/aws/eks/"+clustername+"/cluster", "kube-apiserver-audit", nil)
```
Kaizhe, line 126: aws supports two filter per log group which is not many, it's probably a good idea to support using the existing filter as an option.

plugins/k8saudit-eks/pkg/k8sauditeks/k8sauditeks.go
Issif: what env vars are you thinking about?

plugins/k8saudit-eks/pkg/k8sauditeks/k8sauditeks.go
Issif: I don't understand your point, the idea here is to create a nice UX, the users just set the name of their cluster and the filter is generated by the plugin.

plugins/k8saudit-eks/pkg/k8sauditeks/k8sauditeks.go
Kaizhe: [AWS_WEB_IDENTITY_TOKEN_FILE](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html)

Issif: technically, it will work, because I'm using the official SDK, if the env var is there, it will be used

plugins/k8saudit-eks/pkg/k8sauditeks/k8sauditeks.go
Issif: Ok, I didn't used a subscription, so we're not limited in number of filters per log group. Should be ok.

plugins/k8saudit-eks/pkg/k8sauditeks/k8sauditeks.go
Kaizhe: sgtm

plugins/k8sauditeks/README.md
jasondellaluce: You should be all set with the new feature we introduced in Falco with https://github.com/falcosecurity/falco/pull/2190. It will be part of the upcoming Falco 0.33.0 release.

In your case, you should attain what you want by editing the official k8saudit ruleset with something like:
```yaml
- required_plugin_versions:
  - name: k8saudit
    version: 0.1.0
    alternatives:
      - name: k8saudit-eks
        version: 0.1.0
```

Issif: > You should be all set with the new feature we introduced in Falco with [falcosecurity/falco#2190](https://github.com/falcosecurity/falco/pull/2190). It will be part of the upcoming Falco 0.33.0 release.

I agree, it should work as is, I tried to respect

> In your case, you should attain what you want by editing the official k8saudit ruleset with something like:
> 
> ```yaml
> - required_engine_version: 15
> - required_plugin_versions:
>   - name: k8saudit
>     version: 0.1.0
>     alternatives:
>       - name: k8saudit-eks
>         version: 0.1.0
> ```

Ok, let me update the README to explain that

plugins/k8saudit-eks/rules/k8s_audit_rules.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+- rule: Dummy rule
+  desc: >
+    Dummy rule
+  condition: >
+    ka.verb in (get,create,delete,update)
+  output: user=%ka.user.name verb=%ka.verb target=%ka.target.name target.namespace=%ka.target.namespace resource=%ka.target.resource
+  priority: WARNING
+  source: k8s_audit
+  tags: [k8s]
```
leogr, line 31: Could we remove the `rules/` folder from here?

(otherwise, the CI will automatically publish a package for that)

plugins/k8saudit-eks/go.mod
```diff
@@ -0,0 +1,19 @@
+module github.com/falcosecurity/plugins/plugins/k8saudit-eks
+
+go 1.19
+
+require (
+	github.com/falcosecurity/plugin-sdk-go v0.5.0
+	github.com/falcosecurity/plugins/plugins/k8saudit v0.0.0-20220824115709-c23dc2a4657e
```
leogr, line 7: We should use a tagged version here.

leogr (commented):
Left some comments.

We also need to add an entry into `registry.yaml`, but we can do it also in a follow-up PR.

plugins/k8saudit-eks/go.mod
Issif: I used the 0.3.0 :+1:

Issif:
> Left some comments.
> 
> We also need to add an entry into `registry.yaml`, but we can do it also in a follow-up PR.

We agreed with @jasondellaluce to merge this merge first, then create a second one with the Changelog, the first tag and the update of the registry

jasondellaluce:
> @leogr
We agreed with @jasondellaluce to merge this merge first, then create a second one with the Changelog, the first tag and the update of the registry

Of those, I think I would add the plugin to the registry right away even if not officially released/tagged, just with the purpose of assigning and reserving a plugin ID. If we want to wait for the first real release, I would suggest to open an issue after this PR gets merged to remind us that this plugin still needs to get an official ID assigned.

Issif:
> > @leogr
> > We agreed with @jasondellaluce to merge this merge first, then create a second one with the Changelog, the first tag and the update of the registry
> 
> Of those, I think I would add the plugin to the registry right away even if not officially released/tagged, just with the purpose of assigning and reserving a plugin ID. If we want to wait for the first real release, I would suggest to open an issue after this PR gets merged to remind us that this plugin still needs to get an official ID assigned.

Ok, I reserved the ID with my last commit

jasondellaluce:
> Ok, I reserved the ID with my last commit https://github.com/falcosecurity/plugins/pull/134/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R61
You should edit `registry.yaml`, not the readme (it is auto generated)

Issif:
> > Ok, I reserved the ID with my last commit [#134 (files)](https://github.com/falcosecurity/plugins/pull/134/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R61)
> > You should edit `registry.yaml`, not the readme (it is auto generated)

Sorry, I did it, and changed the name to respect the naming convention enforces by the Makefile

plugins/k8saudit_eks/Makefile
```diff
@@ -0,0 +1,35 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+#
+
+SHELL=/bin/bash -o pipefail
+GO ?= go
+
+NAME := k8saudit_eks
```
leogr, line 17: I prefer k8saudit-eks. 

We prefer to use underscores only in source names. The dash is preferable for plugin names (I know, dummy_c is unfortunate 😅)

Issif: @leogr I prefer the dash too, but I followed the current convention, only the underscore is allowed, see:
https://github.com/falcosecurity/plugins/blob/master/build/registry/pkg/registry/validate.go#L24

```go
var (
	rgxName   = regexp.MustCompile(`^[a-z]+[a-z0-9-_]*$`)
	rgxSource = regexp.MustCompile(`^[a-z]+[a-z0-9_]*$`)
)
```

Issif: Ok, I know, I had to rebase on the upstream, in my locale version, the dash wasn't allows by the validate pkg, see:
```shell
❯ make check-registry
error: plugin name does follow the naming convention: 'k8saudit-eks'
```

After a rebase, it's ok.

leogr: :+1:

leogr (approved):

PR #136, opened 2022-08-11
feat: parse SNS originating from S3

### LINKED ISSUE #135: cloudtrail plugin: Support S3-based SNS ingestion
**Motivation**

A common practice is to use org-wide cloudtrails. These trails write events to s3 nested under account number. As it stands today, there is no easy way to selectively ingest only specific accounts.

**Feature**

We would like to selectively ingest cloudtrail data from aws accounts in an org-trail s3 bucket by way of configuring s3 bucket notifications with prefixes and pointing them to SNS.  Currently the plugin expect SNS notifications directly from cloudtrail.

**Alternatives**

**Additional context**

Looking at the [snsMessage](https://github.com/falcosecurity/plugins/blob/master/plugins/cloudtrail/pkg/cloudtrail/source.go#L71) struct, it appears that the plugin will only ingest directly from cloudtrail. The proposal is to introduce a flag/param that can be used to direct the plugin to read SNS originating from s3.

adnankobir:
**What type of PR is this?**

/kind feature

/area plugins

Adds support for cloudtrail plugin to pull cloudtrail logs from s3 based on s3-event SNS. Further details in issue.  

Fixes #135 

### DIFF
--- a/plugins/cloudtrail/pkg/cloudtrail/source.go
+++ b/plugins/cloudtrail/pkg/cloudtrail/source.go
@@ -265,13 +265,15 @@ func getMoreSQSFiles(pCtx *Plugin, oCtx *PluginInstance) error {
 
 		for _, record := range s3Event.Records {
 
-			// init s3 and only re-init if bucket changes
+			// init s3 and set bucket changes
 			if !s3Init || record.S3.Bucket.Name != lastBucket {
 				oCtx.s3.bucket = record.S3.Bucket.Name
 
-				initS3(oCtx)
-
-				s3Init = true
+				// only init s3 once
+				if !s3Init {
+					initS3(oCtx)
+					s3Init = true
+				}
 			}
 
 			isCompressed := strings.HasSuffix(record.S3.Object.Key, ".json.gz")
--- a/plugins/cloudtrail/pkg/cloudtrail/source.go
+++ b/plugins/cloudtrail/pkg/cloudtrail/source.go
@@ -29,6 +29,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/aws/aws-lambda-go/events"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/sqs"
 	"github.com/aws/aws-sdk-go-v2/service/sqs/types"
@@ -246,6 +247,42 @@ func getMoreSQSFiles(pCtx *Plugin, oCtx *PluginInstance) error {
 
 	if messageType.(string) != "Notification" {
 		return fmt.Errorf("received SQS message that was not a SNS Notification")
+	}
+
+	if pCtx.Config.UseS3SNS {
+		// Process SNS message coming from S3
+		var (
+			s3Event    events.S3Event
+			s3Init     bool
+			lastBucket string
+		)
+
+		err = json.Unmarshal([]byte(sqsMsg["Message"].(string)), &s3Event)
+
+		if err != nil {
+			return err
+		}
+
+		for _, record := range s3Event.Records {
+
+			// init s3 and only re-init if bucket changes
+			if !s3Init || record.S3.Bucket.Name != lastBucket {
+				oCtx.s3.bucket = record.S3.Bucket.Name
+
+				initS3(oCtx)
+
+				s3Init = true
+			}
+
+			isCompressed := strings.HasSuffix(record.S3.Object.Key, ".json.gz")
+
+			oCtx.files = append(oCtx.files, fileInfo{name: record.S3.Object.Key, isCompressed: isCompressed})
+
+			lastBucket = record.S3.Bucket.Name
+		}
+
+		return nil
+
 	}
 
 	var notification snsMessage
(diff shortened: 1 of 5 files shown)


jasondellaluce:
cc @ldegio @leogr @Issif

plugins/cloudtrail/pkg/cloudtrail/source.go
```diff
@@ -248,6 +249,42 @@ func getMoreSQSFiles(pCtx *Plugin, oCtx *PluginInstance) error {
...
+			return err
+		}
+
+		for _, record := range s3Event.Records {
+
+			// init s3 and only re-init if bucket changes
+			if !s3Init || record.S3.Bucket.Name != lastBucket {
+				oCtx.s3.bucket = record.S3.Bucket.Name
+
+				initS3(oCtx)
```
jasondellaluce, line 272: Do you think it's possible for `initS3` to be invoked multiple times? Is it safe?

jasondellaluce (commented):
Haven't tested it, but I think this is a good improvement.

My only concern is whether this new feature should be part of the open parameters (e.g. a new prefix such as `s3-sns://`), or leave it in the init config. My preference would be the first option, as technically this only affect how the event source generates events, and not the plugin as a whole.

Looking forward to hear the opinion of the others too!

leogr:
> My only concern is whether this new feature should be part of the open parameters (e.g. a new prefix such as `s3-sns://`), or leave it in the init config. My preference would be the first option, as technically this only affect how the event source generates events, and not the plugin as a whole.

I think that option should not be part of open parameters, since it does not affect the content of the events.

mstemm:
Hi @adnankobir thanks for the contribution!

I tried to set this up by creating Event notifications for an s3 bucket and copying cloudtrail logs to the s3 bucket. I had a falco instance running subscribing to the sqs queue. The falco instance does get new messages as new files are copies to the s3 bucket, but the format of the message is different than the messages falco gets when all of sns + sqs + cloudtrail are configured.
```
{
    "Records": [
        {
            "s3": {
                "object": {
                    "sequencer": "006303C412AAA08989",
                    "eTag": "f3f98be0210bded2eacd547e208e6562",
                    "size": 2205228,
                    "key": "cloudtrail-4.json"
                },
                "bucket": {
                    "arn": "arn:aws:s3:::mstemm-falco-testing",
                    "ownerIdentity": {
                        "principalId": "xxx"
                    },
                    "name": "mstemm-falco-testing"
                },
                "configurationId": "all-object-create-events-sqs",
                "s3SchemaVersion": "1.0"
            },
            "responseElements": {
                "x-amz-id-2": "AbXgukDlNKEnIQEk8Ua15vu4S9d2dYdn1TBXXqnnY4whd0g6taQ0+0QVA1ZfGGPpAexg+Y3GKyo/0Lh1Q4T+k0RnIrcg5R8/",
                "x-amz-request-id": "7Y0B1QNZ2EP4A1YD"
            },
            "requestParameters": {
                "sourceIPAddress": "xxx"
            },
            "userIdentity": {
                "principalId": "AWS:xxx"
            },
            "eventName": "ObjectCreated:Put",
            "eventTime": "2022-08-22T17:59:52.491Z",
            "awsRegion": "us-east-1",
            "eventSource": "aws:s3",
            "eventVersion": "2.1"
        }
    ]
}
```

How did you set up the bucket, cloudtrail, sns, sqs, etc? I was generally following the instructions here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/ways-to-add-notification-config-to-bucket.html

plugins/cloudtrail/pkg/cloudtrail/source.go
adnankobir: Looking through the function `initS3`, it initiates a new AWS session and creates a new downloader with this session. The intention behind this was as follows:
- The new event notification from S3 includes the [bucket](https://github.com/aws/aws-lambda-go/blob/main/events/s3.go#L40) details.
- If the bucket changes, set `oCtx.s3.bucket` to reflect the new bucket

So I believe its safe to initiate new sessions, but you are correct in identifying this - since I made the assumption that we'd need to init a new session to change the bucket, which is not the case, as we can simply re-use the same session to access the new bucket. Will updated the PR to reflect this change. Thank you!

adnankobir:
> How did you set up the bucket, cloudtrail, sns, sqs, etc? I was generally following the instructions here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/ways-to-add-notification-config-to-bucket.html

Hello @mstemm I've got the bucket, cloudtrail, sns and sqs setup as follows:
```
organization cloudtrial -> s3 -> bucket-notifications on subfolder (ie 'AWSLogs/{org-id}/{account-id}/') -> SNS -> SQS
```

> I tried to set this up by creating Event notifications for an s3 bucket and copying cloudtrail logs to the s3 bucket. I had a falco instance running subscribing to the sqs queue. The falco instance does get new messages as new files are copies to the s3 bucket, but the format of the message is different than the messages falco gets when all of sns + sqs + cloudtrail are configured.

The messages are indeed different than what you would get directly from cloudtrail, for this reason, I am unmarshaling to  [S3Event](https://github.com/aws/aws-lambda-go/blob/main/events/s3.go#L12) and then proceeding to extract the same `fileInfo` as before. At which point, it is business as usual for falco. 

Did you build a new `libcloudtrail.so` based on this PR and set `useS3SNS` to `true`?

mstemm:
Ok that's what I thought I had set up. I started over from scratch today and my falco instance is getting notifications from the s3 bucket directly and processing any new log files as they arrive. So I believe is this all working as expected. Approved!

mstemm (approved):

PR #150, opened 2022-09-20
build: do not include README.md in packages

leogr:
**What type of PR is this?**

/kind cleanup

/kind design

/area build

This is required because some tools (ie. falcoctl or other custom scripts) may use the `.tar.gz` to install a plugin into the default location (ie, `/usr/share/falco/plugins`. In such cases, the README.md file should be not installed.
Removing it from the packages will avoid the need of implementing custom logic for that across different tools.

Fixes #


jasondellaluce (approved):
/approve

PR #151, opened 2022-09-21
fix k8saudit rule EphemeralContainers Created

hi120ki:
Signed-off-by: Hi120ki <12624257+hi120ki@users.noreply.github.com>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area plugins

> /area registry

> /area build

> /area documentation

I'm using kubernetes on GKE, and deploy k8saudit based on admission controller.

In k8saudit detection, a rule `EphemeralContainers Created` is not works because of `kmodify` condition.
`kmodify` condition is

```yaml
- macro: kmodify
  condition: (ka.verb in (create,update,patch))
```

But when creating ephemeral container by `kubectl debug` command, `ka.verb` is null.

> When create Pod, kubernetes audit log records `methodName: "io.k8s.core.v1.pods.create"` and generate `ka.verb: create`.
> But when creating ephemeral container by `kubectl debug` command, kubernetes audit log records `methodName: "io.k8s.core.v1.pods.ephemeralcontainers.patch"` and `ka.verb` is null.

So, I delete condition `kmodify` and it works well.

I'm waiting for your comments on my proposed rules.

Thank you.

Fixes #

### DIFF
--- a/plugins/k8saudit/rules/k8s_audit_rules.yaml
+++ b/plugins/k8saudit/rules/k8s_audit_rules.yaml
@@ -295,7 +295,7 @@
 - rule: EphemeralContainers Created
   desc: >
     Detect any ephemeral container created
-  condition: kevt and pod_subresource and kmodify and ka.target.subresource in (ephemeralcontainers) and not user_known_pod_debug_activities
+  condition: kevt and pod_subresource and ka.target.subresource in (ephemeralcontainers) and not user_known_pod_debug_activities
   output: Ephemeral container is created in pod (user=%ka.user.name pod=%ka.target.name resource=%ka.target.resource ns=%ka.target.namespace ephemeral_container_name=%jevt.value[/requestObject/ephemeralContainers/0/name] ephemeral_container_image=%jevt.value[/requestObject/ephemeralContainers/0/image])
   priority: NOTICE
   source: k8s_audit


leogr:
cc @darryk10

darryk10:
Hi @hi120ki 
Have you tried with self-managed clusters? this might be GKE dealing with k8audit logs in a different way. 
Could you please test this in different environments?
Thanks

AlbertoPellitteri:
Hello! 
I tested the `EphemeralContainers created` rule in a couple of K8s clusters deployed with **kOps** or **kubeadm**. The rule triggered as expected employing the kmodify macro.
So as @darryk10 said, the problem might be related to the fact that GKE audit logs format is different.

hi120ki:
I see. Thank you for inspection. I understand this issue is related to the GKE audit logs format.
Can I close this issue (no merge) because of gke specification?

darryk10:
Hi @hi120ki, since the plugin is related to k8saudit i think we should keep the rule as it is. 
In case there would be a specific plugin for GKE this change would come useful so thanks for your contribution :)

leogr:
> Can I close this PR (no merge) because of gke specification?

:+1: 

We will probably need a dedicated flavor of the `k8saudit` plugin for GKE (ie. `k8saudit-gke`)

hi120ki:
Thank you for your great support.

I would be happy to contribute when `k8saudit-gke` is released!

PR #156, opened 2022-09-21
add new fields with count of failed and denied MFA requests

Issif:
Signed-off-by: Thomas Labarussias <issif+github@gadz.org>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area plugins

> /area registry

> /area build

> /area documentation

This PR introduces 2 new fields to the Okta plugin, it keeps in cache a count of failed and denied MFA pushes by user, it allows to trigger events when too much events occur.

Fields:
|              NAME               |   TYPE   |      ARG      |             DESCRIPTION              |
|---------------------------------|----------|---------------|--------------------------------------|
| `okta.mfa.failure.countlast`    | `uint64` | Key, Required | Count of MFA failures n last seconds |
| `okta.mfa.deny.countlast`       | `uint64` | Key, Required | Count of MFA denies in last seconds  |

Rules:
```yaml
- rule: Too many wrong OTP in last 5min
  desc: An user entered too many wrong OTP in last 5min
  condition: okta.mfa.failure.countlast[300]>=3
  output: "A user has entered too many wrong OTP in last 5min (user='%okta.actor.name' reason=%okta.reason countLast300=%okta.mfa.failure.countlast[300])"
  priority: WARNING
  source: okta
  tags: [okta]

- rule: Too many denied MFA Pushes in last 5min
  desc: An user denied too many MFA Pushes in last 5min
  condition: okta.mfa.deny.countlast[300]>=3
  output: "A user has denied too many MFA pushes in last 5min (user='%okta.actor.name' reason=%okta.reason countLast300=%okta.mfa.deny.countlast[300])"
  priority: WARNING
  source: okta
  tags: [okta]
```

Result:
```
12:18:54.456000000: Warning A user has entered too many wrong OTP (user='Thomas Labarussias' reason=INVALID_CREDENTIALS countLast300=3)
```

Fixes #


plugins/okta/pkg/okta/okta.go
```diff
@@ -343,6 +370,52 @@ func (oktaPlugin *Plugin) Extract(req sdk.ExtractRequest, evt sdk.EventReader) e
...
+					oldest := time.Now().Add(-time.Duration(shift) * time.Second).UnixNano()
+					for _, i := range values {
+						v, _ := strconv.ParseUint(i, 10, 64)
+						if v > uint64(oldest) {
+							count++
+						}
+					}
+				}
+			}
+			req.SetValue(count)
```
leogr, line 394: I think there's an issue here.

This code path is entered each time the field is extracted. The engine calls the field extraction each time it's present in a rule condition or in an output field.

So for example, if two rules use `okta.mfa.failure.countlast`, the counter will be incremented twice each round.
That's not the expected behavior, I guess.

leogr (commented):
This is a great improvement :star_struck: however I believe I spotted a problem. See my comment.

plugins/okta/pkg/okta/okta.go
Issif: I think it's ok, the global count of events is set in cache and this cache is updated only once, at first extraction, see https://github.com/falcosecurity/plugins/pull/156/files#diff-9980b0fcfd1006e1a8087ec9ea20e4cc6f80d1a63dc15da73c2bb5126b497303R250
When we want to count the number of events in last X seconds, we get the cache and increment the `count` variable for each element after (now - X seconds).

plugins/okta/rules/okta_rules.yaml
```diff
@@ -198,3 +198,19 @@
   source: okta
   tags: [okta]
   enabled: false
+
+- rule: Too many failed MFA in last 5min
```
leogr, line 202:
```suggestion
- rule: Too many failed MFA in the last 5min
```

leogr (commented):
Some other comments, after a deeper look

plugins/okta/pkg/okta/okta.go
```diff
@@ -227,6 +235,23 @@ func (oktaPlugin *Plugin) Extract(req sdk.ExtractRequest, evt sdk.EventReader) e
...
+			now := time.Now().UnixNano()
+			valueList := []string{}
+			value, err := oktaPlugin.cache.Get(key)
+			if err == nil {
+				if value != nil {
+					valueList = strings.Split(value.(string), ",")
+				}
+			}
+			valueList = append(valueList, fmt.Sprintf("%v", now))
+			err = oktaPlugin.cache.SetWithExpire(key, strings.Join(valueList, ","), time.Duration(oktaPlugin.CacheExpiration)*time.Second)
```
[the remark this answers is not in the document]
Issif: I had same idea, but I didn't know how to dynamically find the value. If you know, I'll be happy to change. I'll just add few seconds the expiration by security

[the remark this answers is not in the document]
jasondellaluce: The same method can be used to cap the value of `valueList` (its max length should always be `oktaPlugin.maxShift`) at runtime.

Also, here we're using `req.ArgKey()` all over the place, but since we have a positive integer parameter I would declare the field arg with `IsIndex: true`, so that `req.ArgIndex()` can do all the parsing for us out of the box (and the framework will return formatting errors automatically).

jasondellaluce:
Is there a good reason to encode the list of timestamps as a string? This caching library accepts `interface{}` values, so you could simply store a slice of `time.Time` and save the string serialization/deserialization cost.

Moreover, since the time values are ordered in the list by nature, cropping the list length is just a matter of cropping the slice head until we find a time that fits in the interested time window (now - max shift).

plugins/okta/README.md
```diff
@@ -99,6 +103,8 @@ Only `init` accepts settings:
       init_config:
         organization: myorg
         api_token: xxxxxxxxxxx
+        cache_expiration: 84600 #24h
+        cache_maxsize: 84600 #24h
```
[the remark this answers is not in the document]
Issif: fixed

plugins/okta/pkg/okta/okta.go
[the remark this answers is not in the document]
Issif: We can't have a first round with 0 as expiration, the key disappears immediately. I wasn't able to find a nice way to dynamically set up the TTL for keys, so I kept the parameter to configure `cache_expiration`, with a default value to 24h.
The global cache is capped to 10000 keys (hardcoded, I think it's enough and even too big), and now, with the `cache_usermaxsize` we cap the length of each keys.
We need to keep in mind the goal is to store the failed or denied MFA requests, it's useless to store too much of them and for too long, because the idea is to trigger an event when too much occurrences happen in a short period of time, as it might be related to an attempt to hack.

Issif:
> Is there a good reason to encode the list of timestamps as a string? This caching library accepts `interface{}` values, so you could simply store a slice of `time.Time` and save the string serialization/deserialization cost.
> 

You're right, I switched to use directly an array `[]uint64` because it's the format of timestamps.

plugins/okta/pkg/okta/okta.go
```diff
@@ -207,6 +214,8 @@ func (oktaPlugin *Plugin) Fields() []sdk.FieldEntry {
 		{Type: "string", Name: "okta.target.group.id", Desc: "Target Group ID"},
 		{Type: "string", Name: "okta.target.group.aternateid", Desc: "Target Group Alternate ID"},
 		{Type: "string", Name: "okta.target.group.name", Desc: "Target Group Name"},
+		{Type: "uint64", Name: "okta.mfa.failure.countlast", Desc: "Count of MFA failures in last seconds", Arg: sdk.FieldEntryArg{IsRequired: true, IsKey: true}},
+		{Type: "uint64", Name: "okta.mfa.deny.countlast", Desc: "Count of MFA denies in last seconds", Arg: sdk.FieldEntryArg{IsRequired: true, IsIndex: true}},
```
leogr, line 218: The first field is `IsKey` but the latter is `IsIndex`

Is that perhaps a mistake?

plugins/okta/rules/okta_rules.yaml
```diff
@@ -198,3 +198,19 @@
...
+  desc: An user failed MFA too many times in the last 5min
+  condition: okta.mfa.failure.countlast[300]>=3
+  output: "A user has failed MFA to many times in last 5min (user='%okta.actor.name' last_reason='%okta.reason' countLast300=%okta.mfa.failure.countlast[300])"
```
leogr, line 205:
```suggestion
  output: "A user has failed MFA too many times in last 5min (user='%okta.actor.name' last_reason='%okta.reason' countLast300=%okta.mfa.failure.countlast[300])"
```

plugins/okta/pkg/okta/okta.go
Issif: total mistake, thanks.

leogr (approved):

PR #184, opened 2022-10-21
fix(plugins/k8saudit): Recursively parse the files in a directory

### LINKED ISSUE #182: k8saudit plugin - Log File Backend, recursively check the directory
**Describe the bug**

The k8s audit plugin doesn't recursively search the directory for log files.

**How to reproduce it**

For retention purposes, configure k8s to save audit logs as files. Modify the values.yaml of to mount the volume from the host to /var/log/k8saudit on the Falco containers. Configure the plugin's open_params to point to the container audit directory.

**Expected behaviour**

According to the Falco book (O'Reilly, 2022, page 153), for log files the config should be with ">":
open_params: >
/file/path

The greater than (>) sign includes the carriage return (\n) which is an invalid symbol for the Golang URL parser. It worked when I just typed the URL on the same line as the key.

Falco opens the directory, recursively searches the contents and parses each audit file it finds.

**Evidence**

```
pod/falco-fusion-falco-66b76889c4-47lzk
Tue Oct 18 08:08:00 2022: Falco version 0.32.2
Tue Oct 18 08:08:00 2022: Falco initialized with configuration file /etc/falco/falco.yaml
Tue Oct 18 08:08:00 2022: Loading plugin (k8saudit) from file /usr/share/falco/plugins/libk8saudit.so
Tue Oct 18 08:08:00 2022: Loading plugin (json) from file /usr/share/falco/plugins/libjson.so
Tue Oct 18 08:08:00 2022: Configured rules filenames:
Tue Oct 18 08:08:00 2022:    /etc/falco/falco_rules.yaml
Tue Oct 18 08:08:00 2022:    /etc/falco/falco_rules.local.yaml
Tue Oct 18 08:08:00 2022: Loading rules from file /etc/falco/falco_rules.yaml:
Tue Oct 18 08:08:00 2022: Loading rules from file /etc/falco/falco_rules.local.yaml:
Tue Oct 18 08:08:00 2022: Watching /etc/falco/falco.yaml
Tue Oct 18 08:08:00 2022: Watching /etc/falco/falco_rules.yaml.
Tue Oct 18 08:08:00 2022: Watching /etc/falco/falco_rules.local.yaml.
Tue Oct 18 08:08:00 2022: Starting internal webserver, listening on port 8765
Events detected: 0
Rule counts by severity:
Triggered rules by rule name:
Syscall event drop monitoring:
   - event drop detected: 0 occurrences
   - num times actions taken: 0
Error: read /var/log/k8saudit: is a directory
```

**Environment**
- Falco version:
0.32.2
- System info:
23~20.04.2-Ubuntu SMP
- Cloud provider or hardware configuration:
- OS:
NAME="Ubuntu"
VERSION="20.04.5 LTS (Focal Fossa)"
- Kernel:
Linux testlab-0 5.15.0-1017-gcp 23~20.04.2-Ubuntu SMP x86_64 
- Installation method:
Helm chart
**Additional context**
None

AleksVAnd:
/kind bug
/area plugins

Currently, the plugin tries to open any string which is provided to open_params. This succeeds for filepaths but fails if it is point to a directory. 

This PR adds the logic to check if the input string refers to a single file or to a directory.

Fixes #182 

It would be cool if this can be tagged with hacktoberfest

### DIFF
--- a/plugins/k8saudit/pkg/k8saudit/source.go
+++ b/plugins/k8saudit/pkg/k8saudit/source.go
@@ -50,11 +50,22 @@ func (k *Plugin) Open(params string) (source.Instance, error) {
 	case "https":
 		return k.OpenWebServer(u.Host, u.Path, true)
 	case "": // by default, fallback to opening a filepath
-		file, err := os.Open(params)
-		if err != nil {
-			return nil, err
+		// string contains a file extension => it is a single file
+		if strings.Contains(params, ".") {
+			file, err := os.Open(params)
+			if err != nil {
+				return nil, err
+			}
+			return k.OpenReader(file)
+		} else {
+			filepath.Walk(params, func(path string, info os.FileInfo, err error) error {
+				if ! info.IsDir() {
+					// auditFile := info.Name()
+					k.OpenReader(info)
+				}
+				return nil, err
+			})
 		}
-		return k.OpenReader(file)
 	}
 
 	return nil, fmt.Errorf(`scheme "%s" is not supported`, u.Scheme)
--- a/plugins/k8saudit/pkg/k8saudit/source.go
+++ b/plugins/k8saudit/pkg/k8saudit/source.go
@@ -27,6 +27,7 @@ import (
 	"os"
 	"strings"
 	"time"
+	"path/filepath"
 
 	"github.com/falcosecurity/plugin-sdk-go/pkg/sdk"
 	"github.com/falcosecurity/plugin-sdk-go/pkg/sdk/plugins/source"
--- a/plugins/k8saudit/pkg/k8saudit/source.go
+++ b/plugins/k8saudit/pkg/k8saudit/source.go
@@ -27,7 +27,6 @@ import (
 	"os"
 	"strings"
 	"time"
-	"path/filepath"
 
 	"github.com/falcosecurity/plugin-sdk-go/pkg/sdk"
 	"github.com/falcosecurity/plugin-sdk-go/pkg/sdk/plugins/source"
@@ -38,6 +37,12 @@ const (
 	webServerShutdownTimeoutSecs = 5
 	webServerEventChanBufSize    = 50
 )
+
+type AuditMap struct {
+	fileName string
+	modified time.Time // TODO - sort by date, oldest first?
+	pointer io.Reader
+}
 
 func (k *Plugin) Open(params string) (source.Instance, error) {
 	u, err := url.Parse(params)
@@ -52,17 +57,42 @@ func (k *Plugin) Open(params string) (source.Instance, error) {
 		return k.OpenWebServer(u.Host, u.Path, true)
 	case "": // by default, fallback to opening a filepath
 		trimmed := strings.TrimSpace(params)
-		filepath.Walk(trimmed, func(path string, info os.FileInfo, err error) error {
-			if ! info.IsDir() {
-				auditFile := filepath.Join(trimmed, info.Name())
-				file, err := os.Open(auditFile)
+
+		if strings.HasSuffix(trimmed, ".log") {
+			return k.OpenReader(trimmed)
+		}
+
+		files, err := ioutil.ReadDir(trimmed)
+		if err != nil {
+			return nil, err
+		}
+
+		// pointer to the struct to be able to modify it afterwards
+		k8saudit := make(map[int]*AuditMap)
+		numFiles := 0
+
+		for _, f := range files {
+			fmt.Println(f.Name())
+			if !f.IsDir() && strings.HasSuffix(f.Name(), ".log") {
+				var auditFile io.Reader
+				auditFile, err := os.Open(f.Name())
 				if err != nil {
 					return nil, err
 				}
-				k.OpenReader(file)
+				logFile := &AuditMap{fileName: f.Name(), modified: f.ModTime(), pointer: auditFile}
+				k8saudit[numFiles] = logFile
+				numFiles += 1
 			}
-			return nil, err
-		})
+		}
+
+		results := []io.Reader
+		for i := 0; i < len(k8saudit); i++ {
+			results = append(results, k8saudit[i].pointer)
+		}
+		// results... = type []any
+		AllAuditFiles := io.MultiReader(results...)
+		return k.OpenReader(AllAuditFiles)
+
 	}
 
 	return nil, fmt.Errorf(`scheme "%s" is not supported`, u.Scheme)
--- a/plugins/k8saudit/pkg/k8saudit/source.go
+++ b/plugins/k8saudit/pkg/k8saudit/source.go
@@ -27,6 +27,7 @@ import (
 	"os"
 	"strings"
 	"time"
+	"sort"
 
 	"github.com/falcosecurity/plugin-sdk-go/pkg/sdk"
 	"github.com/falcosecurity/plugin-sdk-go/pkg/sdk/plugins/source"
@@ -74,7 +75,7 @@ func (k *Plugin) Open(params string) (source.Instance, error) {
 		})
 
 		// open all files as reader
-		results := []io.Reader
+		results := []io.Reader{}
 		for _, f := range files {
 			if !f.IsDir() {
 				auditFile, err := os.Open(f.Name())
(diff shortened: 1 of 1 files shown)


jasondellaluce (commented):
Thanks for you contribution! At first sight, I don't think this could work, because the return value of `OpenReader` gets discarded! I think we should either create a "multi-file" reader implementation, or implement an "OpenFileList" (bad naming, I know) method. Did you test this?

AleksVAnd:
> 



> 

Hey, Jason,

What do you think: if I add a map[string]File{} which contains the name of the audit log files (keys) and the file descriptors (values)? Then range through the map and invoke OpenReader() for each one?

jasondellaluce:
Within the `Open` function, helpers such as `OpenReader` must be called just once and their return value should be passed through. The return value is an interface `source.Instance`, which is actually the one used by the plugin framework to generate the event stream. 

So, I think that the way to go is to:
- Find a way to open multiple files within the same `OpenXXX` function. This can either mean re-implementing  a new method from scratch or reuse `OpenReader` by masking the multi-file logic behind an `io.Reader`
- Keep watching for new files while generating events. Saving a static map of the directory files might be not enough, as new files can be added while we read the previous ones

AleksVAnd:
> Within the `Open` function, helpers such as `OpenReader` must be called just once and their return value should be passed through. The return value is an interface `source.Instance`, which is actually the one used by the plugin framework to generate the event stream.
> 
> So, I think that the way to go is to:
> 
> * Find a way to open multiple files within the same `OpenXXX` function. This can either mean re-implementing  a new method from scratch or reuse `OpenReader` by masking the multi-file logic behind an `io.Reader`
> * Keep watching for new files while generating events. Saving a static map of the directory files might be not enough, as new files can be added while we read the previous ones

I think I managed to do 1/2. Not sure how to go do "keep watching for new files" - please suggest.

plugins/k8saudit/pkg/k8saudit/source.go
```diff
@@ -50,11 +56,47 @@ func (k *Plugin) Open(params string) (source.Instance, error) {
 	case "https":
 		return k.OpenWebServer(u.Host, u.Path, true)
 	case "": // by default, fallback to opening a filepath
-		file, err := os.Open(params)
+		trimmed := strings.TrimSpace(params)
+
+		if strings.HasSuffix(trimmed, ".log") {
```
jasondellaluce, line 61:
```suggestion
		fileInfo, err := os.Stat(trimmed)
		if err != nil {
			return nil, err
		}
		if !fileInfo.IsDir() {
```
The `.log` suffix is arbitrary and it can be restricting. Let's just check that the file is not a directory.

plugins/k8saudit/pkg/k8saudit/source.go
```diff
@@ -50,11 +56,47 @@ func (k *Plugin) Open(params string) (source.Instance, error) {
...
+			}
+		}
+
+		results := []io.Reader
+		for i := 0; i < len(k8saudit); i++ {
+			results = append(results, k8saudit[i].pointer)
+		}
+		// results... = type []any
+		AllAuditFiles := io.MultiReader(results...)
+		return k.OpenReader(AllAuditFiles)
```
jasondellaluce, line 98: This block can be simplified as follows:
```suggestion
		// sort files by modificationt time
		sort.Slice(files, func(i, j int) bool {
			return files[i].ModTime().Before(files[j].ModTime())
		})

		// open all files as reader
		results := []io.Reader
		for _, f := range files {
			if !f.IsDir() {
				auditFile, err := os.Open(f.Name())
				if err != nil {
					return nil, err
				}
				results = append(results, auditFile)
			}
		}

		// concat all readers into a multi-reader
		return k.OpenReader(io.MultiReader(results...))
```

Also, be mindful that `k.OpenReader` will expect an `io.ReadCloser`, which the result of `io.MultiReader` does not implement. Did you test that this works?

plugins/k8saudit/pkg/k8saudit/source.go
```diff
@@ -38,6 +38,12 @@ const (
 	webServerEventChanBufSize    = 50
 )
 
+type AuditMap struct {
```
jasondellaluce, line 41: Looking at the suggestions below, this block is not necessary anymore.

jasondellaluce (commented):
> I think I managed to do 1/2. Not sure how to go do "keep watching for new files" - please suggest.

I realize this is something complex to achieve after attempting it locally. It's feasible, but would need some extra work that is out of scope for this contribution. Perhaps, we may want to have a `tail://` open param prefix to follow file changes in the future so that we can support the rotating logging backend of the API server. I left you a review.

plugins/k8saudit/pkg/k8saudit/source.go
AleksVAnd: Hey, Jason, I thought that the file names would have the ".log" extension by default, but I understand that you may not want to limit it in such a way. I used `io.NopCloser()` to wrap the readers so it should satisfy `k.OpenReader`. 
So far, I have only tested this logic on dummy JSONL files - tomorrow I can test it in an actual cluster.

AleksVAnd: One thing that worries me is that when files are merged, the MultiReader concatenates the last JSON object with the first JSON object from the next file which results in two objects on the same line. This breaks the JSONL convention and may cause Falco to error during parsing. 
`{"year":"2018","host":"Russia","winner":"France"}
{"year":"2014","host":"Brazil","winner":"Germany"}
{"year":"2010","host":"South Africa","winner":"Spain"}
{"year":"2006","host":"Germany","winner":"Italy"}{"year":"2018","host":"Russia","winner":"France"}
{"year":"2014","host":"Brazil","winner":"Germany"}
{"year":"2010","host":"South Africa","winner":"Spain"}
{"year":"2006","host":"Germany","winner":"Italy"}`

plugins/k8saudit/pkg/k8saudit/source.go
```diff
@@ -50,11 +51,44 @@ func (k *Plugin) Open(params string) (source.Instance, error) {
...
 	case "": // by default, fallback to opening a filepath
-		file, err := os.Open(params)
+		trimmed := strings.TrimSpace(params)
+
+		fileInfo, err := os.Stat(trimmed)
+		if err != nil {
+			return nil, err
+		}
+		if !fileInfo.IsDir() {
+			file, err += os.Open(trimmed)
```
jasondellaluce, line 61:
```suggestion
			file, err := os.Open(trimmed)
```
Typo

plugins/k8saudit/pkg/k8saudit/source.go
```diff
@@ -50,11 +51,44 @@ func (k *Plugin) Open(params string) (source.Instance, error) {
...
+
+		sort.Slice(files, func(i, j int) bool {
+			return files[i].ModTime().Before(files[j].ModTime())
+		})
+
+		// open all files as reader
+		results := []io.Reader{}
+		for _, f := range files {
+			if !f.IsDir() {
+				auditFile, err := os.Open(f.Name())
```
jasondellaluce, line 81:
```suggestion
				auditFile, err := os.Open(trimmed + "/" + f.Name())
```
We need to add the base file path, otherwise this will always fail (just tested it locally).

plugins/k8saudit/pkg/k8saudit/source.go
```diff
@@ -50,11 +51,44 @@ func (k *Plugin) Open(params string) (source.Instance, error) {
...
+				if err != nil {
+					return nil, err
+				}
+				results = append(results, auditFile)
```
jasondellaluce, line 85:
```suggestion
				results = append(results, auditFile)
				results = append(results, strings.NewReader("\n"))
```
Very simple but yet functioning solution to the issue you pointed out! Great catch btw.

jasondellaluce:
> Hey, Jason, I thought that the file names would have the ".log" extension by default, but I understand that you may not want to limit it in such a way.

You're right, but also consider that this feature can be useful for test purposes too. For example, with `make test-files` we can already pull a bunch of test files with `.json` extension pulled from the Falco regression tests, and thanks to this feature it's now very easy to test the plugin agains all of them!

jasondellaluce (approved):
/approve

Thanks!

PR #192, opened 2022-12-14
update(build/registry) change the logic how new versions of artifacts are discovered

alacuku:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area plugins

/area registry

/area build

> /area documentation

This PR changes the source of truth used to discover new versions of plugins and rulesfiles. The CI parses the `registry.yaml` file and for each entry:
  * Plugins: checks if the plugin entry has already been pushed to the
    OCI registry. If found, retrieves the latest version otherwise
    do nothing. It checks if there have been released new versions
    of the plugin by checking the release tags in the local git repo.
    If yes, it searches and downloads the new released versions from
    the s3 bucket and pushes them to the OCI repo.
  * Rules: since the rules doe not have an entry in the registry.yaml
    file it checks if the plugins has a rulesfile url. If yes does the
    same thing as for the plugins but for the rulesfile associated to
    the plugin entry.

Furthermore, it adds the floating tags based on the major and minor number
of the semver for each artifact. The floating tags are needed by #falcosecurity/falcoctl#189

Fixes #

### DIFF
--- a/build/registry/cmd/registry/registry.go
+++ b/build/registry/cmd/registry/registry.go
@@ -23,6 +23,7 @@ import (
 	"github.com/falcosecurity/falcoctl/pkg/oci/authn"
 	"github.com/falcosecurity/plugins/build/registry/pkg/registry"
 	"github.com/falcosecurity/plugins/build/registry/pkg/registry/distribution"
+	"github.com/falcosecurity/plugins/build/registry/pkg/registry/oci"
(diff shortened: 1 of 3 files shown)


leogr (approved):
:+1: 

SGTM. Let's try :)

PR #205, opened 2022-12-28
feat(plugins/k8saudit): add header-based custom fields to events

### LINKED ISSUE #206: k8s-audit: Add custom fields to audit events
**Motivation**

Instead of running a falco + falcosidekick instance for each of our self-hosted k8s clusters to add extra metadata like the k8s cluster name to the events, I would prefer to run a central falco cluster capable of ingesting k8s audit events from all of our clusters. Maintaining a per cluster falco instance becomes cumbersome.

**Feature**

Some way to add additional/extra/custom fields to each k8s audit event, just like falcosidekick can do, based on metadata sent with the request to the k8saudit plugin. As we are limited by what the k8s webhook audit backend can do, I am thinking of using http headers to supply additional metadata. This setup allows running a 'centralised' falco (cluster) ingesting k8s audit events from multiple k8s clusters without loosing for example the origin k8s cluster name.

Hopefully the custom fields added by the k8saudit plugin can be picked up by the solution for [falco #2127](https://github.com/falcosecurity/falco/issues/2127)

**Alternatives**

Currently running a falco instance for each k8s cluster, with a falcosidekick instance next to each falco instance. Falco ingests the k8s audit events, and pushes them to the falcosidekick instance. This falcosidekick instance adds the additional cluster specific labels before forwarding events to another falcosidekick instance used for falcosidekick ui and forwarding to another system. This setup becomes quite complex to maintain.

**Additional context**

As described in [falco #2289](https://github.com/falcosecurity/falco/issues/2289):

My goal is to run a central falco (cluster) ingesting k8saudit events (or other cloud based events supported by falco plugins). A k8saudit event though misses important metadata to be able to trace the event back to the cluster (https://github.com/falcosecurity/falco/issues/1704), like a cluster name/id.

The only way I see to do this, is by using the client authentication to provide additional metadata. A (reverse)proxy/ingress controller is responsible for handling authentication and pass additional metadata based on the authentication to falco by setting additional http headers. This is usually a pretty standard feature of a proxy (e.g. X-Forward-For headers). For example we can use mtls with k8s cluster specific client certs to authenticate. The proxy can forward the certificate subject field to the k8saudit plugin, and we can encode the cluster name in the subject for example.

I am looking into extending the k8saudit plugin with support to grab additional fields from http headers. Passing these values along the plugin event processing pipeline gets a bit ugly though. Only option seems to be altering the event from the http body with the header metadata. This unfortunately seems to mean the field names are fixed, as required by the extractor part.
Last challenge is to actually get the extra fields in the output, apart from editing all rules, the -p options seems atm the only way (related https://github.com/falcosecurity/falco/issues/1704#issuecomment-1028185613). See https://github.com/falcosecurity/falco/issues/2127 as well for additional output fields.
Downside of this approach is that each plugin requires a dedicated falco cluster, as the -p flag (with %field markers) is specific to each plugin. Falco validates the -p argument very early on, even before excluding rules with the -T/-t flags. E.g. adding a '%ka.auditid' with -p fails even if you only enable k8s rules (-t k8s).

Another downside is that the extra metadata is added to the output field. A better alternative is to only add these fields to the 'output_fields', not the 'output'. Just like falcosidekick allows to set 'custom fields', which are only added to the 'output_fields'.
Furthermore it would be ideal to be able to add 'custom' named fields to the output_fields of the alert. The additional fields a user wants to set, based on authentication/headers, is completely up to the user. Especially if you are forwarding the alerts with falcosidekick to other tools, the extra output_fields basically become 'labels' standardised across the organisation. E.g. alerts are forwarded as log lines to a log aggregation system. Each log line has a number of fixed labels so its origin can easily be found. The kubernetes cluster of origin is just one possibility, but one could add labels like division, project or team.

sboschman:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area plugins

> /area registry

> /area build

> /area documentation

For context see:
- [falco #2289](https://github.com/falcosecurity/falco/issues/2289)
- [falco #1704](https://github.com/falcosecurity/falco/issues/1704)
- [falco #2127](https://github.com/falcosecurity/falco/issues/2127)

Fixes #206

Sorry for the spam, tried to open a 'draft pr', and hit the button to change pr type, ~but I guess that isn't supported by github on this repository~ nvm I am blind, found the sneaky little convert to draft link , The code needs a cleanup before a review. Just wanted a draft pr to get some feedback on the idea/concept and the plugin configuration.

### DIFF
--- a/plugins/k8saudit/README.md
+++ b/plugins/k8saudit/README.md
@@ -13,6 +13,10 @@ The plugin implementation is a 1-1 porting of the legacy implementation supporte
 
 This plugin supports consuming Kubernetes Audit Events coming from the [Webhook backend](https://kubernetes.io/docs/tasks/debug/debug-cluster/audit/#webhook-backend) or from file. For webhooks, the plugin embeds a webserver that listens on a configurable port and accepts POST requests. The posted JSON object comprises one or more events. The webserver of the plugin can be configuted as part of the plugin's init configuration and open parameters. For files, the plugins expects content to be [in JSONL format](https://jsonlines.org/), where each line represents a JSON object, containing one or more audit events.
 
+The Kubernetes webhook audit backend configuration is basically a `kubeconfig`. This does not allow setting custom HTTP headers to sent along with the POST requests. However, it does allow defining authentication by username+password or certificate.
+A reverse proxy in front of the plugin embedded webserver for authentication allows for passing authentication info as 'forwarded' headers. Based on the headers passed by a reverse proxy this plugin can add custom fields to each event. For example the id/name of the originating Kubernetes cluster of the event, as the Kubernetes audit event itself does not contain cluster identification.
+The custom fields are not part of the defined supported extraction fields. To use custom fields in rules a json reference has to be used, e.g. `%json.value[/cluster_name]`
+
 The expected way of using the plugin is through Webhook. The file reading support is mostly designed for testing purposes and for development, but does not represent a concrete deployment use case.
 
 ## Capabilities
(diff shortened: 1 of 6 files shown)


sboschman:
Hello @jasondellaluce and @leogr , as github automatically added you as reviewers, hopefully you can spare a moment to give some feedback on the idea/concept behind this PR. In the first place, is this feat a useful addition to the k8s-audit plugin at all, or do you see a better approach to add additional metadata to k8s audit events. And secondly, is the new `customFieldsHeaders` config option looking oke.

The actual code is not ready for a code review yet, but feel free to have a peek at it if that would help understanding the approach I am taking.

The readme snippet I added to express the idea behind the config option I added:
> The Kubernetes webhook audit backend configuration is basically a `kubeconfig`. This does not allow setting custom HTTP headers to sent along with the POST requests. However, it does allow defining authentication by username+password or certificate.
A reverse proxy in front of the plugin embedded webserver for authentication allows for passing authentication info as 'forwarded' headers. Based on the headers passed by a reverse proxy this plugin can add custom fields to each event. For example the id/name of the originating Kubernetes cluster of the event, as the Kubernetes audit event itself does not contain cluster identification.
The custom fields are not part of the defined supported extraction fields. To use custom fields in rules a json reference has to be used, e.g. `%json.value[/cluster_name]`.

The actual config yaml:

```yaml
plugins:
  - name: k8saudit
    library_path: libk8saudit.so
    init_config:
      sslCertificate: /etc/falco/falco.pem
      customFieldsHeaders:
        # Example HTTP header sent with the request:
        # X-WebAuth-User: cheese-cluster
        - header: "X-WebAuth-User"
          customFields:
            - cluster_name
        # Example HTTP header sent with the request:\
        # X-Forwarded-TlS-Client-Cert-Info: Subject=\"DC=org,DC=cheese,OU=CheeseFactory,OU=Packaging,CN=*.example.com\"
        - header: "X-Forwarded-TlS-Client-Cert-Info"
          pattern: "Subject=\"DC=[A-Za-z0-9]+,DC=[A-Za-z0-9]+,OU=([A-Za-z0-9]+),OU=([A-Za-z0-9]+),.+\""
          customFields:
            - division
            - team
```
  
Another added snippet from the readme to describe the new config options:
>**Initialization Config**:
>- `customFieldsHeaders`: List of custom fields to add to every audit event based on HTTP request headers (Optional)

>**HTTP Header-based custom fields Config**
>- `header`: HTTP header name to extract custom field(s) from (Required)
>- `pattern`: Regex pattern containing capture groups, in order each capture group corresponds to a custom field (Optional)
>- `customFields`: List of custom fields to add, without `pattern` only a single custom field can be defined (Required)

sboschman:
/remove-lifecycle stale

Just waiting for feedback on the idea of adding custom fields based on headers and if such a feature would be accepted, before spending more time on the code.

This feature should be seen as part of a larger effort to build a centralised Kubernetes audit log system with Falco. In which multiple k8s clusters sent their audit logs to this central system for analysis and further processing (with the Falco stack). In such a setup we need a way to enrich the incoming k8s audit logs with additional metadata not present in the audit log messages themselves. E.g. the name of the originating k8s cluster.

PR #207, opened 2023-01-03
feat(build/registry): add config layer to artifacts when pushing to registry

alacuku:
Signed-off-by: Aldo Lacuku <aldo@lacuku.eu>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area plugins

/area registry

/area build

> /area documentation

The CI before pushing the artifacts to the remote registry it correctly configures the `config` layer by adding dependencies and requirements. 

The `rulesfiles` could depend on `plugins` meaning that they need a specific version of a plugin in order to be consumed. At the same time they require that the  Falco `engine version` needs to satisfy the minimum version specified in the `rulesfile`.   The `config` layer of the `k8saudit` ruleset artifact:
```yaml
{
  "name": "k8saudit-rules",
  "version": "0.4.1",
  "dependencies": [
    {
      "name": "json",
      "version": "0.3.0"
    },
    {
      "name": "k8saudit",
      "version": "0.1.0",
      "alternatives": [
        {
          "name": "k8saudit-eks",
          "version": "0.1.0"
        }
      ]
    }
  ],
  "requirements": [
    {
      "name": "engine_version",
      "version": "15.0.0"
    }
  ]
}
```

On the other hand the `plugins` require that the Falco instance loading them satisfies the `plugin_api_version`. The following snippet show the `config` layer for the `k8saudit` plugins:
```yaml
{
  "name": "k8saudit",
  "version": "0.4.1",
  "requirements": [
    {
      "name": "plugin_api_version",
      "version": "2.0.0"
    }
  ]
}
```

The `config` layer is used by the `falcoctl` to manage dependencies and requirements when installing/updating artifacts falcosecurity/falcoctl#205

Fixes #

### DIFF
--- a/build/registry/pkg/oci/dependencies.go
+++ b/build/registry/pkg/oci/dependencies.go
@@ -1,0 +1,82 @@
+/*
+Copyright (C) 2022 The Falco Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package oci
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/falcosecurity/falcoctl/pkg/oci"
+	"gopkg.in/yaml.v3"
+)
+
+const depsKey = "- required_plugin_versions"
+
+// ErrDepNotFound error when the dependencies are not found in the rulesfile.
+var ErrDepNotFound = errors.New("dependencies not found")
+
+// rulesfileDependencies given a rulesfile in yaml format it scans it nad extracts its dependencies.
+func rulesfileDependencies(fileName string) ([]oci.ArtifactDependency, error) {
+	var start bool
+	var buf []byte
+	var deps []oci.ArtifactDependency
+
+	// Open the file.
+	file, err := os.Open(fileName)
+	if err != nil {
+		return nil, fmt.Errorf("unable to open file %q: %v", fileName, file)
+	}
+
+	// Prepare the file to be read line by line.
+	fileScanner := bufio.NewScanner(file)
+	fileScanner.Split(bufio.ScanLines)
+
+	// Is appended to each line when inserted in the buffer.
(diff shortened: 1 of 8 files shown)


build/registry/pkg/oci/dependencies.go
```diff
@@ -0,0 +1,82 @@
...
+	fileScanner := bufio.NewScanner(file)
+	fileScanner.Split(bufio.ScanLines)
+
+	// Is appended to each line when inserted in the buffer.
+	newLine := []byte("\n")
+
+	// Falco rulesfiles are a list of dictionaries. We only want the "required plugin versions" by the ruleset. We do
+	// not want to load all the file in memory, so we scan it line by line. When we reach the interested section we save
+	// each line in a buffer, and after that we unmarshal it to a proper data structure.
+	for fileScanner.Scan() {
```
jasondellaluce, line 57: I'm not a big fan of this reading loop, I would much prefer doing the YAML deserialization first and then searching for the right key after. However the Falco ruleset YAML files are very loosely typed and can't be parsed easily in any case, so I see the reasons behind this solution.

Let's keep this approach but set a remainder that we may want to find a more elegant solution in the future.

build/registry/pkg/oci/requirements.go
```diff
@@ -0,0 +1,126 @@
...
+		return nil, fmt.Errorf("unable to open file %q: %v", filePath, file)
+	}
+
+	defer file.Close()
+
+	// Prepare the file to be read line by line.
+	fileScanner := bufio.NewScanner(file)
+	fileScanner.Split(bufio.ScanLines)
+
+	for fileScanner.Scan() {
```
jasondellaluce, line 82: Same as comment above

build/registry/pkg/oci/requirements.go
```diff
@@ -0,0 +1,126 @@
...
+	}, nil
+}
+
+// pluginRequirement given a plugin as a shared library it loads it and gets the api version
+// required by the plugin.
+func pluginRequirement(filePath string) (*oci.ArtifactRequirement, error) {
+	cPath := C.CString(filePath)
+	var cErr *C.char
+
+	handler := C.pluginOpen(cPath, &cErr)
```
jasondellaluce, line 112: Friendly reminder that the latest plugin SDK go release introduced a `loader` package that abstracts all the plugin loading logic by using the same code as Falco (https://github.com/falcosecurity/plugin-sdk-go/blob/main/pkg/loader/loader.go).

This should save us from dealing with all the CGO madness in many different places. However, in most tools under the `/build` directory of this repo we haven't adopted the plugin loader yet, so I'd suggest refactoring all the tools in a future PR.

jasondellaluce (approved):
/approve

PR #227, opened 2023-02-02
Sort k8saudit data model.

distortedsignal:
**What type of PR is this?**

/kind cleanup

/kind documentation

/area plugins

/area documentation

Organize the K8s Audit plugin datatypes. I think this is needed because it helps users see the relationships between various fields and the objects that contain them. I was trying to see how various things related to each other, and it was hard to do because all the fields were unordered.

N/A

I don't think this removed any fields, but if it did, please feel free to ask me to re-add them.


distortedsignal:
🙄 \@ whatever bot added the Size/M tag to this PR.

distortedsignal:
boop @ whoever needs to look at this to get it in.

jasondellaluce:
Just saw this. Please not that the list is auto-generated in the readmes, so you either:
- reorder the fields in the plugin's code, or
- update the readme generation tool to sort the fields by name before generating the MD table

Option 1 is not feasibile in the general case, because the order of fields matters in the plugins code (their index can be used in the piece of logic responsible of extracting them from the events). However, I think we don't do that in the k8saudit plugin specifically. I think option 2 is the one that makes more sense, but it depends on your ordering needs. Let me know what you think.

distortedsignal:
> ...I think option 2 is the one that makes more sense, but it depends on your ordering needs...

Ok, that's reasonable.

Could you point me to where the code is that generates this documentation?

Thanks.

jasondellaluce:
Here: https://github.com/falcosecurity/plugins/blob/master/build/readme/fields.go

I think you could just sort the fields once they are read from the plugin.

distortedsignal:
> ...
> I think you could just sort the fields once they are read from the plugin.

I was thinking the same thing. Unfortunately, the last time I worked on this PR was a while ago, and I don't know when I'll get the next chance to. I'll try to get to it this month.

PR #232, opened 2023-02-17
update k8s registry domain

cpanato:
**What type of PR is this?**
/kind cleanup

/area plugins

- update k8s registry domain

part of https://github.com/kubernetes/k8s.io/issues/4780

/assign @leogr 

Fixes #

### DIFF
--- a/plugins/k8saudit/rules/k8s_audit_rules.yaml
+++ b/plugins/k8saudit/rules/k8s_audit_rules.yaml
@@ -186,6 +186,7 @@
   tags: [k8s]
 
 # These container images are allowed to run with hostnetwork=true
+# TODO: Remove k8s.gcr.io reference after 01/Dec/2023
 - list: falco_hostnetwork_images
   items: [
     gcr.io/google-containers/prometheus-to-sd,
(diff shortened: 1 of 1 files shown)


leogr (commented):
Since the rules can also run on environments not up-to-date, I'd keep both domains (ie `k8s.grc.io` and `registry.k8s.io`).

@LucaGuerra @darryk10 @loresuso wdyt?

darryk10:
I agree on keeping them both. Some k8s version will still remain in the old domain so it makes sense to whitelist them both. In the future we would consider removing the old one.

cpanato:
sound good and thanks for the feedback, i will update those and add a comment
@leogr @darryk10

plugins/k8saudit/rules/k8s_audit_rules.yaml
```diff
@@ -186,6 +186,7 @@
   tags: [k8s]
 
 # These container images are allowed to run with hostnetwork=true
+# TODO: Remove k8s.gcr.io reference after 01/Dec/2023
```
cpanato, line 189: @leogr @darryk10 added this date to we remove the k8s reference after December 
the k8s registry will be frozen on 3rd of April 2023 (https://kubernetes.io/blog/2023/02/06/k8s-gcr-io-freeze-announcement/) so we give some extra time for people to update the clusters.
wdyt?

leogr (approved):
LGTM

/hold for a second opinion cc @falcosecurity/rules-maintainers @falcosecurity/plugins-maintainers

LucaGuerra:
/unhold

LGTM, thanks!

PR #238, opened 2023-02-28
new(ci): add release pipelines

cappellinsamuele:
**What type of PR is this?**

/kind feature

/area build

This PR introduces two GHA pipelines that replicate the behavior of the prow jobs that are currently responsible for the release of the plugin builds on dev and stable channels.
The final goal is the disposal of the previously mentioned prow jobs, which shall be replaced with these pipelines.

`NONE`

`NONE`


.github/workflows/update_registry.yml
FedeDP, line 1: I have mixed feelings about this one; same as https://github.com/falcosecurity/rules/pull/36.
I am not sure we want to push directly onto `falcoctl` gh-pages branch from a gha. I am not even sure if that would work from a public fork...
@leogr wdyt?

PR #242, opened 2023-03-16
feat(plugins/k8saudit): include query params in health check url macros

### LINKED ISSUE #241: k8saudit: Adjust healthcheck macros to include query parameters
**Motivation**

The included rule "Anonymous Request Allowed" excludes request made to healthcheck endpoints. But the exception does not cover requests made to healthcheck endpoints with query parameters like `?exclude=kms-provider-0`. This leads to a lot of findings.

**Feature**

Add exception for calls that start with `/healthz?` and so on.

**Alternatives**

Do it locally only.

**Additional context**

Here is macro definition:

https://github.com/falcosecurity/plugins/blob/904f750adab96bb958e4647c1984bc98a240023a/plugins/k8saudit/rules/k8s_audit_rules.yaml#L147-L154

Here is only usage of macros:

https://github.com/falcosecurity/plugins/blob/904f750adab96bb958e4647c1984bc98a240023a/plugins/k8saudit/rules/k8s_audit_rules.yaml#L266-L273

trallnag:
**What type of PR is this?**

/kind feature

/area plugins

Extends macros that match call to Kubernetes control plane health check endpoints to include query parameters. This is supported by Kubernetes officially, see [link](https://kubernetes.io/docs/reference/using-api/health-checks/).

Fixes #241

### DIFF
--- a/plugins/k8saudit/rules/k8s_audit_rules.yaml
+++ b/plugins/k8saudit/rules/k8s_audit_rules.yaml
@@ -145,13 +145,13 @@
   condition: ka.target.resource=secrets
 
 - macro: health_endpoint
-  condition: ka.uri=/healthz
+  condition: ka.uri=/healthz or ka.uri startswith /healthz?
 
 - macro: live_endpoint
-  condition: ka.uri=/livez
+  condition: ka.uri=/livez or ka.uri startswith /livez?
 
 - macro: ready_endpoint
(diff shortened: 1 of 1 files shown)


leogr:
Thank you!

/ok-to-test

cc @darryk10 @loresuso

leogr (approved):

PR #255, opened 2023-04-04
feature(plugins/cloudtrail): add fields to plugin

SKosier:
**What type of PR is this?**

/kind feature

/area plugins

Fixes #


plugins/cloudtrail/pkg/cloudtrail/extract.go
```diff
@@ -500,6 +503,30 @@ func getfieldStr(jdata *fastjson.Value, field string) (bool, string) {
 			return false, ""
 		}
 		res = iname
+	case "ec2.imageid":
+		var imageId = ""
+		jilist := jdata.GetArray("responseElements", "tagSpecificationSet", "items")
+		if jilist == nil {
```
jasondellaluce, line 509:
```suggestion
		if jilist == nil || len(jilist) == 0 {
```

It's very unlikely that we get a malformed payload, but being cautions is always better than panic-ing at runtime :)

jasondellaluce (approved):
/approve

PR #272, opened 2023-04-20
fix(ci): added permissions for S3 upload and volume mount on qemu.

cappellinsamuele:
**What type of PR is this?**

/kind bug

/area build

This PR adds permissions to `push_master` and `release` GHA to allow authentication on S3. 
It also fixes volume the volume mounts in arm64 builds.

`NONE`

`NONE`

### DIFF
--- a/.github/workflows/push_master.yml
+++ b/.github/workflows/push_master.yml
@@ -51,9 +51,13 @@ jobs:
         with:
           arch: aarch64
           distro: ubuntu_latest
+          githubToken: ${{ github.token }}
+
+          setup: |
+            mkdir -p "${PWD}/artifacts"
+
           dockerRunArgs:
             --volume "${PWD}/artifacts:/artifacts"
-          githubToken: ${{ github.token }}
  
           install: |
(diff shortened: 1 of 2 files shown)


.github/workflows/push_master.yml
```diff
@@ -47,14 +51,18 @@ jobs:
         with:
           arch: aarch64
           distro: ubuntu_latest
+          dockerRunArgs:
```
FedeDP, line 54: We also need t ocreate the folder in the `setup` step: 
```
 setup: |
            mkdir -p "${PWD}/artifacts"
```
See https://github.com/uraimo/run-on-arch-action#advanced-example

.github/workflows/release.yml
```diff
@@ -63,13 +67,16 @@ jobs:
           arch: aarch64
           distro: ubuntu_latest
           githubToken: ${{ github.token }}
-
+          dockerRunArgs:
```
FedeDP, line 70: As above.

jasondellaluce (approved):

PR #278, opened 2023-05-10
fix(ci): properly use golang:1.18 image, that was the image used before, with the correct glibc version

FedeDP:
**What type of PR is this?**

/kind bug

/area build

Moreover, fixed CI artifacts versioning.

Fixes #


.github/workflows/push_master.yml
```diff
@@ -19,10 +25,21 @@ env:
 jobs:
   release-plugins-dev-x86:
     runs-on: ubuntu-latest
+    container: golang:1.18
```
FedeDP, line 28: This is the same container we were using on prow.

.github/workflows/push_master.yml
```diff
@@ -19,10 +25,21 @@ env:
...
     steps:
+      - name: Install deps
+        run: |
+          apt update
+          apt install -y --no-install-recommends git awscli
+    
       - name: Checkout Plugins ⤵️
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
FedeDP, line 38: Fixes artifacts versioning.

.github/workflows/push_master.yml
```diff
@@ -19,10 +25,21 @@ env:
...
 
+      - name: Safe directory
```
FedeDP, line 40: Since container image is run as root, we need to set the safe dir.

.github/workflows/push_master.yml
```diff
@@ -58,15 +77,22 @@ jobs:
 
           dockerRunArgs:
             --volume "${PWD}/artifacts:/artifacts"
+            --volume "/var/run/docker.sock:/var/run/docker.sock"
```
FedeDP, line 80: We need to run a docker in docker.

.github/workflows/push_master.yml
```diff
@@ -58,15 +77,22 @@ jobs:
...
  
           install: |
-            apt-get update && apt-get install -y --no-install-recommends build-essential git ca-certificates golang curl && apt-get clean
-            update-ca-certificates
-            git config --global --add safe.directory ${{ github.workspace }}
+            apt-get update && apt-get install -y --no-install-recommends docker.io && apt-get clean
             
           run: |
-            make packages
+            cat > script.sh <<EOF
```
FedeDP, line 86: Create the build script that will be executed by the docker container.

.github/workflows/push_master.yml
```diff
@@ -58,15 +77,22 @@ jobs:
...
+            #!/bin/sh
+            apt-get update && apt-get install -y --no-install-recommends ca-certificates curl && apt-get clean
+            update-ca-certificates
+            git config --global --add safe.directory /repo
+            cd /repo && make packages
             cp -R output/* /artifacts
+            EOF
+            chmod +x script.sh
+            docker run -ti --rm --platform arm64 -v /artifacts:/artifacts -v ${{ github.workspace }}:/repo golang:1.18 /repo/script.sh
```
FedeDP, line 95: Riddle: 
How many levels of docker in docker in docker in docker are we?

.github/workflows/release.yml
FedeDP, line 1: I think a reusable workflow, similar to what i did for Falco, would fit best here.

FedeDP: I can do it after this gets merged and tested (this PR's fix is much more important btw)

jasondellaluce (approved):

PR #361, opened 2023-09-22
fix(plugins/gcp): update extract function

lorenzo-merici:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area plugins

> /area registry

> /area build

> /area documentation

Fixes #


jasondellaluce:
/close

Closing this one after discussing it with the contributor. The rationale is that we don't want to move from the current behavior of returning NULL when a field is missing. The correct semantics of Falco is to check whether a field exists or not with the `exists` operator in case it can return null.

PR #375, opened 2023-11-15
feat(plugins/k8saudit/rules) add detection for portforwarding

### LINKED ISSUE #230: k8s plugins should detect portforwards as well as execs
**Motivation**

Currently the various example k8s alerting rules don't include detections for port-forwards.
Similar to the exec rule(s) there should be an example rule for port-forwards  because that actor could be sidestepping various network segmentation/auth steps.

**Feature**
Add port-forward detection, which should be almost identical to the exec detections

**Alternatives**

**Additional context**

RichardoC:
**What type of PR is this?**

/kind feature

/area plugins

Currently there aren't any detections for port forwarding, which isn't great. Port forwarding can be used to avoid controls such as service meshes, and can be used for data exfiltration. There's rarely a good reason for anyone to be using these regularly so it should have a detection, just like exec already does.

For anyone wondering, you can only port forward a pod. When you kubectl port forward svc/myservice it actually checks the selectors for the service and port mapping, and port forwards one of the pods which matches those selectors, on the required port after mapping through the service. This is why I'm only reporting the pod that is port forwarded, there's no way to know if the user was trying to port forward a pod, or service unless we correlated back to whether this user had done a service lookup and then a pod list before issuing this API call.

Fixes #230 

This hasn't been tested locally, it's based on the Kubernetes API spec so should be tested before merging

### DIFF
--- a/plugins/k8saudit/rules/k8s_audit_rules.yaml
+++ b/plugins/k8saudit/rules/k8s_audit_rules.yaml
@@ -298,6 +298,18 @@
   source: k8s_audit
   tags: [k8s]
 
+- macro: user_known_portforward_activities
+  condition: (k8s_audit_never_true)
+
+- rule: port-forward
+  desc: >
+    Detect any attempt to portforward
+  condition: ka.target.subresource in (portforward) and not user_known_portforward_activities
+  output: Portforward to pod (user=%ka.user.name pod=%ka.target.name ns=%ka.target.namespace action=%ka.target.subresource )
+  priority: NOTICE
+  source: k8s_audit
+  tags: [k8s]
+
 - macro: user_known_pod_debug_activities
   condition: (k8s_audit_never_true)
 


Andreagit97:
Thank you for this, not a big expert here but we will try to take a look ASAP

loresuso:
Hi @RichardoC and thanks for the contribution!
I do agree that this is something we need and NOTICE priority. I think it might be worth it to introduce an allow list also at the namespace level
In any case, LGTM!

RichardoC:
> Hi @RichardoC and thanks for the contribution! I do agree that this is something we need and the NOTICE priority. I think it might be worth it to introduce an allow list also at the namespace level In any case, LGTM!

I could add this, but nothing else in this file is allowlisted by namespace. Wouldn't it be better to have that as a macro for all rules on namespaced objects?

Also, did you have any luck testing these with a real cluster?

loresuso:
That's right, let's keep the namespace allow list for a second PR in case. 
Re testing the rule, no I didn't have the chance to try it so far. We could follow this guide https://falco.org/docs/install-operate/third-party/learning/#temp-qa-integrations-to-minikube to set up Falco with the k8s audit log plugin in a k8s cluster to try it out.

RichardoC:
Good news, after some faffing about got this working on minikube
image

loresuso:
Great news, thanks for trying it out! 
not a maintainer here, @leogr mind approving if you agree?

leogr (approved):
LGTM.


A little disclaimer: I think it's okay to merge this. However, note that the new k8s ruleset will not released soon. We will likely release all new plugin versions near the next Falco release. We can open a follow-up PR if we need to introduce any minor change to this new rule before releasing it (like enforcing the [Rules Maturity Framework](https://github.com/falcosecurity/rules/blob/main/CONTRIBUTING.md#rules-maturity-framework) policies).

cc @falcosecurity/rules-maintainers @falcosecurity/plugins-maintainers

leogr:
> @leogr are you planning to move these rules out to the rules repo?

No, we don't have any plans for that yet. We have just to understand if and how the Rules Maturity Framework applies to plugin rulesets. Eventually, it will have a minor impact on plugin rules (ie. probably we will just add maturity tags).

PR #378, opened 2023-12-01
Add `k8smeta` plugin to the plugin repo

Andreagit97:
**What type of PR is this?**

/kind feature

/area plugins

This PR adds the `k8smeta` plugin to the plugin repo.


Andreagit97:
This PR is still a draft but i need the CI to check the integration of the plugin, so i cannot mark it as a draft

Andreagit97:
The plugin repo could be another great candidate for https://github.com/falcosecurity/falco/pull/2945, since from now for the `k8smeta` plugin we will build grpc on QEMU for arm64 :/  it takes almost 1h and 33 m

Andreagit97:
I have disabled codeQL for `cpp` in the "general" CI job, probably i will add a specific codeQL job for the `k8smeta` plugin and the `dummy_c` one

Andreagit97:
I will split the codeQL job in another PR that should be merged before this

plugins/k8smeta/src/plugin.cpp
```diff
@@ -0,0 +1,1185 @@
...
+    return true;
+}
+
+// Obtain a param from a sinsp event
+static inline sinsp_param get_syscall_evt_param(void* evt, uint32_t num_param)
+{
+    uint32_t dataoffset = 0;
+    // pointer to the lengths array inside the event.
+    auto len = (uint16_t*)((uint8_t*)evt + sizeof(falcosecurity::_internal::ss_plugin_event));
+    for(uint16_t j = 0; j < num_param; j++)
```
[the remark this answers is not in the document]
Andreagit97: this is solved now

plugins/k8smeta/README.md
```diff
@@ -0,0 +1,79 @@
...
+|----------------------------|-----------------|---------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `k8sres.pod.name`          | `string`        | None          | Kubernetes pod name.                                                                                                                                                                                                                                                                                         |
+| `k8sres.pod.id`            | `string`        | None          | Kubernetes pod ID.                                                                                                                                                                                                                                                                                           |
```
incertum, line 25: Update as well to `k8sres.pod.uid`. In general, let;s sync docs between here and the `k8s` fields extracted from the container runtime, see also https://github.com/falcosecurity/libs/pull/1575


During today's maintainer meeting I also didn't quite get why we choose the prefix `k8sres.` -- I think `k8smeta.` would be so much better and make the most sense.

incertum: Additional note for blog post / documentation: Explain very detailed how this differs from the `k8s` fields and when this would be needed or why you should consider this plugin in cases where the `k8s` fields extracted from the runtime socket are all you need.

Andreagit97: thank you for the suggestions about the blog/docs!

Andreagit97: > Update as well to k8sres.pod.uid. In general, let;s sync docs between here and the k8s fields extracted from the container runtime, see also https://github.com/falcosecurity/libs/pull/1575

I agree, i will change it in `*.pod.uid` and same for all others resources, like `namespace.uid`

Andreagit97: > During today's maintainer meeting I also didn't quite get why we choose the prefix k8sres. -- I think k8smeta. would be so much better and make the most sense.

This is a valid point, I don't like too much the name `k8sres`. The idea we talked about with Leo is to express with the prefix what the filter check will extract. in this case `k8sres` means that we will extract a `k8s` resource, probably something like `k8sresource` would be more self-explanatory. BTW `k8smeta` could be a valid name, first is the same name as the plugin and this in some way is coherent and unique, and its meaning could be in line with what we want, `k8smeta` could mean we will extract a `k8s metadata`. It is also in line with the name of the remote collector, it is called `k8s-metacollector`(https://github.com/falcosecurity/k8s-metacollector). WDYT? @leogr

alacuku: I agree that `k8smeta` would be a more suitable name.

leogr: `k8smeta` is ok for me too.

plugins/k8smeta/src/plugin.h
```diff
@@ -0,0 +1,276 @@
...
+};
+
+class my_plugin
+{
+    public:
+    // Keep this aligned with `get_fields`
+    enum K8sFields
+    {
+        K8S_POD_NAME,
+        K8S_POD_ID,
```
leogr, line 63:
```suggestion
        K8S_POD_UID,
```

?

leogr (commented):
Just
- `k8sres` :point_right: `k8smeta`
-  `.*id` :point_right:  `*.uid`

otherwise SGTM! :clap:

plugins/k8smeta/src/plugin.h
Andreagit97: done

plugins/k8smeta/README.md
Andreagit97: solved both points

leogr (approved):

PR #382, opened 2023-12-11
update(dummy_c): create a custom CI workflow for `dummy_c`

Andreagit97:
**What type of PR is this?**

/kind cleanup

/kind design

/area plugins

This change is because we cannot use codeQL autobulid mode for complex cpp plugins (see https://github.com/falcosecurity/plugins/pull/378). So the idea is to create a custom workflow for each cpp plugin (in this case dummy_c) with a `build` job (that runs `codeQL`) and a formatting job.
In the same way, the `k8smeta` plugin will have a dedicated workflow with a `build-and-test` job (that runs `codeQL`) and a formatting job.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,16 @@
+---
+Language: Cpp
+BasedOnStyle: LLVM
+AccessModifierOffset: -8
+BreakBeforeBraces: Allman
+BreakConstructorInitializers: AfterColon
+ColumnLimit: 80
+ConstructorInitializerIndentWidth: 8
+ContinuationIndentWidth: 8
+DerivePointerAlignment: true
+IndentWidth: 4
+SortIncludes: Never
+SpaceAfterTemplateKeyword: false
+SpaceBeforeCtorInitializerColon: false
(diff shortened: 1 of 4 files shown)


.clang-format
Andreagit97, line 1: this file will be used also by the `k8smeta` plugin

.github/workflows/codeql-analysis.yml
Andreagit97, line 1: we cannot rely on the codeQL autobuild mode for complex cpp projects, `dummy_c` was a sort of exception because it is straightforward and doesn't have dependencies, but in order to follow a clear approach, we will have a general codeQL job for `golang` plugins and a dedicated codeQL job for each cpp plugin

.github/workflows/dummy_c-ci.yaml
Andreagit97, line 1: dedicated CI for the dummy_c plugin

plugins/dummy_c/dummy.cpp
Andreagit97, line 1: just applied clang-format on this file

PR #385, opened 2023-12-18
update(registry): add `k8smeta` plugin to the registry

Andreagit97:
**What type of PR is this?**

/kind documentation

/area registry

This PR adds `k8smeta` plugin to the registry

### DIFF
--- a/registry.yaml
+++ b/registry.yaml
@@ -341,3 +341,24 @@ plugins:
         supported: true
         id: 999
         source: test
+  - name: k8smeta
+    description: Enriche Falco syscall flow with Kubernetes Metadata
+    authors: The Falco Authors
+    contact: https://falco.org/community
+    maintainers:
+      - name: The Falco Authors
+        email: cncf-falco-dev@lists.cncf.io
+    keywords:
+      - kubernetes
+      - syscall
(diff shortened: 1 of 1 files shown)


registry.yaml
```diff
@@ -341,3 +341,27 @@ plugins:
...
+      cosign:
+        certificate-oidc-issuer: https://token.actions.githubusercontent.com
+        certificate-identity-regexp: https://github.com/falcosecurity/plugins/
+    capabilities:
+      extraction:
+        supported: true
+      parsing:
+        supported: true
+      async:
+        supported: true
```
Andreagit97, line 367: not clear to me which are the supported values under `capabilities`, do we have a validation for this field?

leogr: I guess we don't have it yet.

leogr: We don't have it yet :point_down: 
https://github.com/falcosecurity/plugins/blob/master/build/registry/pkg/registry/types.go#L39

Could you remove `async` at the moment, please? We can still introduce it later.

Andreagit97: done I used just `extraction` and added the `sources` field. Thank you for the pointer!

PR #401, opened 2024-01-18
fix(ci): fix remote tag deletion/skipping

LucaGuerra:
**What type of PR is this?**

/kind bug

/area build

The remote build would show erratic behavior when a 0.10.0 tag was introduced. This was possibly due to git ordering of tags. Instead of relying on it we'll use a set, remove what we don't need and later sort with semver.

Fixes #

### DIFF
--- a/build/registry/pkg/oci/oci.go
+++ b/build/registry/pkg/oci/oci.go
@@ -227,20 +227,26 @@ func newReleases(artifactName, remoteVersion string) ([]semver.Version, error) {
 		return append(versions, *v), nil
 	}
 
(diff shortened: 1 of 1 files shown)


leogr:
> The remote build would show erratic behavior when a 0.10.0 tag was introduced. This was possibly due to git ordering of tags. Instead of relying on it we'll use a set, remove what we don't need and later sort with semver.

 Make sense. For example:
 
 ```bash
❯ git --no-pager tag --list "okta-[0-9].*" --contains okta-0.8.0
okta-0.10.0
okta-0.8.0
okta-0.8.1
okta-0.9.0
```

leogr (approved):

PR #412, opened 2024-02-12
fix(cloudtrail): fix path and extract more fields from the s3 notification event

matteopasa:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area plugins

> /area registry

> /area build

> /area documentation

Fixes #

### DIFF
--- a/plugins/cloudtrail/pkg/cloudtrail/extract.go
+++ b/plugins/cloudtrail/pkg/cloudtrail/extract.go
@@ -215,6 +215,10 @@ func getfieldStr(jdata *fastjson.Value, field string) (bool, string) {
 	case "ct.id":
 		val := jdata.GetStringBytes("eventID")
 		if val == nil {
+			val = jdata.GetStringBytes("id")
+		}
+
+		if val == nil {
 			return false, ""
 		} else {
 			res = string(val)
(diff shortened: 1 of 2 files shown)


plugins/cloudtrail/pkg/cloudtrail/extract.go
```diff
@@ -214,6 +214,10 @@ func getfieldStr(jdata *fastjson.Value, field string) (bool, string) {
 	switch field {
 	case "ct.id":
 		val := jdata.GetStringBytes("eventID")
+		if val == nil {
+			val = jdata.GetStringBytes("request-id")
+		}
```
Issif, line 219: This situation should never happen, `eventID` is a mandatory field. Moreover, it seems the path is `requestID`:
image

matteopasa: the thing is that the s3 notification event that we are receiving is this one: https://docs.aws.amazon.com/AmazonS3/latest/userguide/ev-events.html 
The part inside the "detail" field, I've tried to map some of the fields to the ct fields but I can remove them

Issif: The plugin is from collecting the audit logs from cloudtrail, we should never receive a direct S3 event. Maybe I'm wrong or there's something I don't get.

matteopasa: I am working on this feature to receive also s3 notification events https://sysdig.atlassian.net/browse/SSPROD-34513
We need to modify the plugin to extract fields also from that kind of events, that is the reason of the change, I spoke with @jasondellaluce and since the change is little he was ok not creating a new plugin or a private plugin for this, but let me know if you agree
The fields we are interested in are the s3 ones, so I can remove the ct ones if they can cause issues

PR #413, opened 2024-02-13
feat: add more Cloudtrail fields

uhei:
Add more fields as defined at
https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-event-reference-record-contents.html

**What type of PR is this?**

/kind feature

/area plugins

The current CloudTrail plugin is missing fields. This PR adds all currently defined fields according to [https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-event-reference-record-contents.html](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-event-reference-record-contents.html).

The fields `ct.addendum.*` and `ct.edgedevicedetails` are untested due to missing events with these fields.

The plugin has only been tested with [falcodump](https://www.wireshark.org/docs/man-pages/falcodump.html), [Logray](https://www.wireshark.org/) and CloudTrail access via S3.


leogr:
cc @Issif

leogr:
cc @Issif @LucaGuerra

Issif (commented):
/lgtm

@leogr do you think we should embed the plugin version bump up in that PR too?

leogr:
> /lgtm
> 
> @leogr do you think we should embed the plugin version bump up in that PR too?

I know there should be other planned changes to this plugin, so we can bump up in a later PR.

Btw, the PR SGTM, too. I will do a second look soon and then approve.

Thanks!

leogr (approved):

PR #417, opened 2024-02-21
feat(cloudtrail): Add ct.response and ct.request field

uhei:
Adding CloudTrail fields `requestParameters` as `ct.request` and `responseElements` as `ct.response`.

**What type of PR is this?**

/kind feature

/area plugins

There are too many parameters/elements in `requestParameters`/`responseElements` for all the different API calls to have a dedicated field for every one. Nevertheless, I want to be able to filter or search for the content of these fields in Logray.

This commit provides the entire content of the two fields.

Maybe we can have a 'Custom CloudTrail requestParameters fields' option in Logray (like we have it for 'Custom HTTP Header Fields') in the future. But this needs to be discussed in a separate issue in the Wireshark project.


leogr:
cc @Issif

Issif:
/lgtm

makes sense to avoid these elements for the ouput

leogr (approved):
:+1:

PR #418, opened 2024-02-23
feat(cloudtrail): Add generic additionalEventData field

uhei:
**What type of PR is this?**

/kind feature

/area plugins

Similar to ct.request, all values should also be available for additionalEventData.


leogr:
/assign

cc @Issif

plugins/cloudtrail/pkg/cloudtrail/extract.go
```diff
@@ -643,6 +644,13 @@ func getfieldStr(jdata *fastjson.Value, field string) (bool, string) {
 		} else {
 			res = string(val)
 		}
+	case "ct.additionaleventdata":
+		val := jdata.Get("additionalEventData")
+		if val == nil {
+			return false, ""
+		} else {
+			res = string(val.MarshalTo(nil))
+		}
```
Issif, line 653:
```suggestion
		}
		res = string(val.MarshalTo(nil))
```

leogr (approved):
LGTM, Thank you!

cc @LucaGuerra

PR #419, opened 2024-02-27
new(anomalydetection): Initial Scope - CountMinSketch Powered Probabilistic Counting and Filtering

incertum:
**What type of PR is this?**

> /kind bug

> /kind cleanup

/kind design

> /kind documentation

> /kind failing-test

/kind feature

/area plugins

> /area registry

> /area build

> /area documentation

Introduce a new `anomalydetection` plugin, as outlined in the [Proposal](https://github.com/falcosecurity/falco/blob/master/proposals/20230620-anomaly-detection-framework.md).

Fixes https://github.com/falcosecurity/falco/issues/3117

https://github.com/falcosecurity/falco/issues/3117

Fixes #

Sharing some hopefully useful notes:

- Check out the README first, as it summarizes current limitations and the initial scope
- Much of the code was duplicated from falcosecurity/libs; I’ve indicated this to show that this code doesn’t need review here. If you find issues, let’s address them in libs instead
- Hot reloading seems to be working (no segfaults observed), except toggling enabled/disabled during a hot reload doesn’t work, which we can likely ignore
- The project builds in a container suitable for building Falco, so we can set up CI
- Initial local tests confirm the intended functionality, including the reset timers
- Unit tests cover the most critical functions but could still be improved
- cms.h contains the algorithm/math (it's probabilistic counting only, not AI/ML); the rest is mostly generic plugin setup or configuration.
- Config initialization is designed to error out if the plugin is used for currently unsupported cases
- The profile field filtercheck code might be removed if libs filterchecks are opened to plugins, which would make it more stable and less risky. Additionally, translating Falco filtercheck logic to the plugin framework was very tedious
- Further testing on more robust servers is needed to check for performance and stability issues, such as CPU spikes or segfaults after extended runs
- Most things had to be wrapped in try-catch statements, as simple checks for nullptr were not possible ...

For this first version, I am primarily concerned about performance and stability (no segfaults). I would truly appreciate a very critical review in this regard. Thanks a bunch in advance!


plugins/anomalydetection/src/plugin.cpp
```diff
@@ -0,0 +1,582 @@
...
+    static_assert(fields_size == ANOMALYDETECTION_FIELD_MAX, "Wrong number of anomaly fields.");
+    return std::vector<falcosecurity::field_info>(fields, fields + fields_size);
+}
+
+bool anomalydetection::extract(const falcosecurity::extract_fields_input& in)
+{
+    auto& req = in.get_extract_request();
+    uint64_t count_min_sketch_estimate = 0;
+    // todo fix settings to allow arg index in field
+    auto index = req.get_arg_index();
```
incertum, line 486: @jasondellaluce may I ask for some guidance re how to make the arg indices available? The idea is to support notations like `anomaly.count_min_sketch[2]` etc. It should be possible, just not sure how. Thanks in advance!

jasondellaluce: In `std::vector<falcosecurity::field_info> anomalydetection::get_fields()` you are defining the list of available fields. In each, when doing `{...}` you're invoking their constructor and defining them one by one. The 5th (and optional) argument of that constructor is the field arg, which is a substruct you can define to represent what the argument for that field looks like: https://github.com/falcosecurity/plugin-sdk-cpp/blob/efa5564221cea590eaf727a7a26e655515848c2b/include/falcosecurity/types.h#L149

Example:
```
...
{
    ft::FTYPE_UINT64,
    "anomaly.count_min_sketch",
    "Count Min Sketch Estimate",
    "Count Min Sketch Estimate according to the specified behavior profile for a predefined set of {syscalls} events. Access different behavior profiles/sketches using indices. For instance, anomaly.count_min_sketch[0] retrieves the first behavior profile defined in the plugins' `init_config`.",
    { // field arg
        false, // key
        true,  // index
        false, // required (don't know, maybe it is -- you know better)
    }
},
...
```

After this declaration, the framework (libs) will handle all the syntax parsing for you and will provide you the user-defined argument, which you can access through `req.get_arg_index()`

incertum: Thank you! I see it's a sub struct :upside_down_face: works like a charm now!

plugins/anomalydetection/src/plugin.cpp
```diff
@@ -0,0 +1,656 @@
...
+        m_pid = m_thread_table.get_field(t.fields(), "pid", st::SS_PLUGIN_ST_INT64);
+        m_ptid = m_thread_table.get_field(t.fields(), "ptid", st::SS_PLUGIN_ST_INT64);
+        m_sid = m_thread_table.get_field(t.fields(), "sid", st::SS_PLUGIN_ST_INT64);
+        m_comm = m_thread_table.get_field(t.fields(), "comm", st::SS_PLUGIN_ST_STRING);
+        m_exe = m_thread_table.get_field(t.fields(), "exe", st::SS_PLUGIN_ST_STRING);
+        m_exepath = m_thread_table.get_field(t.fields(), "exe_path", st::SS_PLUGIN_ST_STRING);
+        m_exe_writable = m_thread_table.get_field(t.fields(), "exe_writable", st::SS_PLUGIN_ST_BOOL);
+        m_exe_upper_layer = m_thread_table.get_field(t.fields(), "exe_upper_layer", st::SS_PLUGIN_ST_BOOL);
+        m_exe_from_memfd = m_thread_table.get_field(t.fields(), "exe_from_memfd", st::SS_PLUGIN_ST_BOOL);
+        // m_args = m_thread_table.get_field(t.fields(), "args", TBD);
```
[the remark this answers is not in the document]
incertum: CC @mrgian

plugins/anomalydetection/src/plugin.cpp
```diff
@@ -0,0 +1,817 @@
...
+            {
+            case PPME_SYSCALL_OPEN_X:
+            case PPME_SOCKET_ACCEPT_X:
+            case PPME_SOCKET_ACCEPT_5_X:
+            case PPME_SOCKET_ACCEPT4_X:
+            case PPME_SOCKET_ACCEPT4_5_X:
+            case PPME_SOCKET_ACCEPT4_6_X:
+            case PPME_SYSCALL_CREAT_X:
+            {
+                // todo get fd and look up the right fd from the table
```
incertum, line 729: @jasondellaluce How do you get the current fd and then do a targeted lookup in the fd table. libs has m_lastevent_fd ... Thanks in advance!

plugins/anomalydetection/test/src/num/cms.ut.cpp
```diff
@@ -0,0 +1,156 @@
...
+    open_inspector();
+
+    int64_t fd = 4;
+    add_event(increasing_ts(), 3, PPME_SYSCALL_OPEN_E, 3, "/tmp/../../the_file", 0, 0);
+    add_event_advance_ts(increasing_ts(), 3, PPME_SYSCALL_OPEN_X, 6, fd, "/tmp/some_other_file", 0, 0, 0, (uint64_t) 0);
+    fd = 5;
+    add_event(increasing_ts(), 3, PPME_SYSCALL_OPEN_E, 3, "/tmp/../../the_file2", 0, 0);
+    evt = add_event_advance_ts(increasing_ts(), 3, PPME_SYSCALL_OPEN_X, 6, fd, "/tmp/some_other_file2", 0, 0, 0, (uint64_t) 0);
+    ASSERT_EQ(get_field_as_string(evt, "fd.name"), "/the_file2");
+    ASSERT_EQ(get_field_as_string(evt, "anomaly.count_min_sketch.profile[1]", pl_flist), "/tmp/../../the_file2/the_file2");
```
incertum, line 155: Currently failing because of https://github.com/falcosecurity/plugins/pull/419/files#r1671669624 @jasondellaluce, thanks!

plugins/anomalydetection/src/plugin.cpp
incertum: We briefly chatted on slack and `lastevent_fd` is indeed not accessible, hence had to write it to the thread_entry within the plugin. It probably makes sense to expand the plugin API a bit further and make those fields accessible as well so that the work does not have to be done twice.

jasondellaluce (approved):
/approve

🥳

Issue #228, opened 2023-02-08
[UMBRELLA] Requested Plugins


Issif:
In January 2022, Falco introduced its first version of a Plugin framework to extend its available inputs. The framework has been enhanced in the following months to have something production ready for adopters.

## Existing Plugins

We, the maintainers of Falco, created a bunch of Plugins to replace deprecated features (k8saudit) or to follow mediatic security events (Okta breach).

Right now, we have [registered](https://github.com/falcosecurity/plugins#registered-plugins) (excluding dummy plugins)

## SDK

To make the development of plugins easier, 2 SDK are provided: [Go](https://github.com/falcosecurity/plugin-sdk-go/) and [C++](https://github.com/falcosecurity/plugin-sdk-cpp). 
We can notice all plugins have been written in Go, it can be explained by several factors:
Go is easier to than C++
It’s a common language in web development, so in adopters’ infras
Falco’s ecosystem already embeds different Go codebases (Falcosidekick, Falcosidekick-UI, Falcoctl, Driverkit, Falco-exporter, Event-generator)

## Libs

Writing a plugin from scratch could be complicated for the contributors, this is why we could also provide libraries to keep them focus on the extraction logic and not the asides (auth, polling, create a web server, etc). The main goal of these libs is to avoid duplicate codes across plugins, allowing to keep an uniformity.

This approach has been started with 2 libs for AWS:
- [AWS Session](https://github.com/falcosecurity/plugins/tree/master/shared/go/aws/session): allows to create easily a session for AWS API
- [AWS Cloudwatch](https://github.com/falcosecurity/plugins/tree/master/shared/go/aws/cloudwatchlogs): allows to set filters and starts the polling of log entries from Cloudwatch

To “open” Falco to more sources, we could create shared libs for generic usages:
- Web Server: to collect JSON webhook payloads
- File reader: to follow new entries in a file
- Kafka: to be a consumer of a topic
- SQS: to poll a queue
- RabbitMQ: to be a consumer of an exchange
- MQTT: for IoT

We also need to address the most common Cloud Providers and their specific log aggregator systems with the basic functions which are:
- Authentication to the API
- Creation of a client for the log service
- Gathering of logs
- Looping over the results

By providing these libs, it will be easier for developers to create new plugins for specific usages with these Cloud Providers.

## Plugins

The purpose of this issue is to list the requested plugins by the community, the volunteers to develop them and their statuses.

The following table will be kept updated to avoid people to search through N issues.

<table>
[...]

PR #424, opened 2024-03-05
A plugin to collect K8S Audit Logs from GKE (k8saudit-gke plugin)

sboschman:
**What type of PR is this?**

/kind feature

/area plugins

This `k8saudit-gke` plugin tries to bridge te gap between Google Cloud audit logs for GKE and the Falco ruleset for k8s audit events. This way the existing Falco ruleset for k8s audit events can be used to monitor GKE clusters as well.

Add a `k8saudit-gke` plugin to the plugins suite (requested plugins are tracked in #228 )

- is it acceptable to bump golang to 1.21 (from 1.18) for plugins?
- default k8s rules have to be fine-tuned a bit for gke (e.g. add the allowed gke images). Currently the rule overrides are in the dir `rule_extensions` as the ci workflow validator fails for override rules. Obviously it would be nice to release/ship this rule file as part of the plugin, so it can be installed with falcoctl. Is this fixable in the ci pipeline?
- the plugin id is currently 666, if we can register a "real" plugin id that would be great


sboschman:
@Issif as I mentioned earlier in [#228](https://github.com/falcosecurity/plugins/issues/228#issuecomment-1972698556) here is the pr for a k8saudit-gke plugin. I did put down some issues/questions in the reviewer notes, some guidance with these would be appreciated.

plugins/k8saudit-gke/pkg/k8sauditgke/config.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/falcosecurity/plugin-sdk-go/pkg/sdk"
+	"github.com/falcosecurity/plugins/plugins/k8saudit/pkg/k8saudit"
+	"github.com/patrickmn/go-cache"
+	"google.golang.org/api/container/v1"
+)
+
+type Plugin struct {
+	k8saudit.Plugin
+	logger *log.Logger
```
Issif, line 31: I see you point to create a logger, but Falco shoud be in charge to print the lines in the stdout, to be sure to have a consistency of the format

plugins/k8saudit-gke/pkg/k8sauditgke/k8sauditgke.go
```diff
@@ -0,0 +1,120 @@
...
+	"strings"
+
+	"github.com/falcosecurity/plugin-sdk-go/pkg/sdk/plugins"
+	"github.com/falcosecurity/plugin-sdk-go/pkg/sdk/plugins/source"
+	"google.golang.org/api/container/v1"
+	"google.golang.org/api/option"
+)
+
+const (
+	PluginID          uint32 = 666
```
Issif, line 33:
```suggestion
	PluginID          uint32 = 16
```

Issif (changes requested):

Issif:
Can you sign off all the commits please.

For the version of Go, @jasondellaluce could help us, last time I tried I got issues, don't remember which ones.

plugins/k8saudit-gke/pkg/k8sauditgke/config.go
sboschman: I guess I must be missing something, but how do you write to the falco logger from the plugin sdk?

Both the [k8saudit](https://github.com/falcosecurity/plugins/blob/master/plugins/k8saudit/pkg/k8saudit/k8saudit.go#L44) and [k8saudit-eks](https://github.com/falcosecurity/plugins/blob/master/plugins/k8saudit-eks/pkg/k8sauditeks/k8sauditeks.go#L47) are creating a plugin logger as well.

The `NewPushInstance` method from the sdk only takes a go channel to pass `PushEvent`s to falco. And the `Plugin` interface does not seem to offer a logger?

Issif: You're right, I forgot that. We should see in the future to extend to API/SDK in Falco for that

sboschman:
> I guess we need to add your plugin here to allow it to use the file:
> 
> https://github.com/falcosecurity/plugins/blob/master/plugins/k8saudit/rules/k8s_audit_rules.yaml#L21,L26

yes indeed, modded my own rules files with:
```
- required_plugin_versions:
  - name: k8saudit
    version: 0.7.0
    alternatives:
      - name: k8saudit-gke
        version: 0.1.0
```

Don't know how the release versioning in the ci workflows does it thing, but I can just add it with version 0.1.0 ?

Issif:
cc @alacuku @LucaGuerra can you help us with the CI for the rules?

leogr:
> This is what happens when I place the gke rule overrides into the `rules` dir ([job](https://github.com/falcosecurity/plugins/actions/runs/8158202437/job/22300120796)):
> 
> ```
> Error: /__w/plugins/plugins/plugins/k8saudit-gke/rules/k8s_audit_rules_gke.yaml: Invalid
> 1 Errors:
> In rules content: (/__w/plugins/plugins/plugins/k8saudit-gke/rules/k8s_audit_rules_gke.yaml:0:0)
>     rule 'Disallowed K8s User': (/__w/plugins/plugins/plugins/k8saudit-gke/rules/k8s_audit_rules_gke.yaml:10:2)
> ------
> - rule: Disallowed K8s User
>   ^
> ------
> LOAD_ERR_YAML_VALIDATE (Error validating internal structure of YAML file): Item has no mapping for key 'output'
> ```
> 
> The rule override:
> 
> ```
> - macro: gke_service_agent
>   condition: (ka.user.name endswith "@container-engine-robot.iam.gserviceaccount.com")
> 
> - rule: Disallowed K8s User
>   condition: and not gke_service_agent
>   override: 
>     condition: append
> ```
> 
> any idea if this is fixable @Issif (or perhaps @jasondellaluce ) ? Or is this kind of plugin rule overrides not allowed by design?

The issue is due because `override` was introduced in Falco 0.37, but the CI is still using Falco 0.36.2.
@alacuku @LucaGuerra 
I guess we can upgrade the CI to use the latest Falco version.

Issif:
> > This is what happens when I place the gke rule overrides into the `rules` dir ([job](https://github.com/falcosecurity/plugins/actions/runs/8158202437/job/22300120796)):
> > ```
> > Error: /__w/plugins/plugins/plugins/k8saudit-gke/rules/k8s_audit_rules_gke.yaml: Invalid
> > 1 Errors:
> > In rules content: (/__w/plugins/plugins/plugins/k8saudit-gke/rules/k8s_audit_rules_gke.yaml:0:0)
> >     rule 'Disallowed K8s User': (/__w/plugins/plugins/plugins/k8saudit-gke/rules/k8s_audit_rules_gke.yaml:10:2)
> > ------
> > - rule: Disallowed K8s User
> >   ^
> > ------
> > LOAD_ERR_YAML_VALIDATE (Error validating internal structure of YAML file): Item has no mapping for key 'output'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The rule override:
> > ```
> > - macro: gke_service_agent
> >   condition: (ka.user.name endswith "@container-engine-robot.iam.gserviceaccount.com")
> > 
> > - rule: Disallowed K8s User
> >   condition: and not gke_service_agent
> >   override: 
> >     condition: append
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > any idea if this is fixable @Issif (or perhaps @jasondellaluce ) ? Or is this kind of plugin rule overrides not allowed by design?
> 
> The issue is because `override` was introduced in Falco 0.37, but the CI still uses Falco 0.36.2. @alacuku @LucaGuerra I guess we can upgrade the CI to use the latest Falco version.

I prepared a PR for https://github.com/falcosecurity/plugins/pull/426

leogr:
> I prepared a PR for #426

#426 is merged now. 

@sboschman could you rebase this, pls? The CI issue should be addressed then. Thanks!

sboschman:
@leogr check now runs with 0.37.1, but we somehow have to tell te checker it has to load the default/base k8saudit rules as well, any ideas on this?

Issif:
> @leogr check now runs with 0.37.1, but we somehow have to tell te checker it has to load the default/base k8saudit rules as well, any ideas on this?

Can you try to add a symlink to the [k8s_audit_rules.yaml](https://github.com/falcosecurity/plugins/blob/ff78f4218758e1dd0d0e4e0a06b7b5d8198a3458/plugins/k8saudit/rules/k8s_audit_rules.yaml), I'm not sure but it could work, because of this https://github.com/falcosecurity/plugins/blob/ff78f4218758e1dd0d0e4e0a06b7b5d8198a3458/.github/setup-plugin-config-rules.sh#L22

sboschman:
@Issif that seems to have solved the `validate-local` job, but the `validate-falcoctl` job still gives an error:
```
Error: rules validation command failed, open /usr/share/falco/plugins/libk8saudit-gke.so: no such file or directory
```

I am not sure what this `Validate plugin and rules` step is doing, is it trying to d/l this new unreleased plugin with falcoctl from the oci registry? In the step script it says something 'echo Attempt installing locally-built plugin "${plugin_name}"...' so should it be able to grab the plugin .so file from the build?

sboschman:
```
    deps=$(cat $rules_file | yq -r '.[].required_plugin_versions | select(. != null and . != "")[] | [.name + ":" + .version] | @csv')
```

I have an idea.. I have no 
```
- required_engine_version: 15
- required_plugin_versions:
```
fields in my rules file

alacuku:
> > @Issif that seems to have solved the `validate-local` job, but the `validate-falcoctl` job still gives an error:
> > ```
> > Error: rules validation command failed, open /usr/share/falco/plugins/libk8saudit-gke.so: no such file or directory
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I am not sure what this `Validate plugin and rules` step is doing, is it trying to d/l this new unreleased plugin with falcoctl from the oci registry? In the step script it says something 'echo Attempt installing locally-built plugin "${plugin_name}"...' so should it be able to grab the plugin .so file from the build?
> > The `plugins-x86_64-424.tar.gz` build artifiact does contain a `k8saudit-gke-0.1.0-0.0.0-0+4cb2951-linux-x86_64` file
> 
> The error simply means that the plugin is not present in the container image. I can't recall the full workflow right now. @LucaGuerra @alacuku, any clue? 🤔

It seems that the tool is installing the wrong plugins: k8saudit plugin instead of the k8saudit-gke..

alacuku:
@sboschman, could you try to remove the rulesfile from the k8saudit-gke plugin? From my understanding, the CI should behave the same way as with k8saudit-eks plugin which shares the same rules file as the k8saudit plugin. The CI creates an empty rulesfile and validates the plugin.

sboschman:
> @sboschman, could you try to remove the rulesfile from the k8saudit-gke plugin? From my understanding, the CI should behave the same way as with k8saudit-eks plugin which shares the same rules file as the k8saudit plugin. The CI creates an empty rulesfile and validates the plugin.

yes that works to pass the ci, but how do we publish the gke specific rule overrides?

plugins/k8saudit/rules/k8s_audit_rules.yaml
```diff
@@ -24,6 +24,8 @@
     alternatives:
       - name: k8saudit-eks
         version: 0.4.0
+      - name: k8saudit-gke
+        version: 0.1.0
```
leogr, line 28: :+1:

plugins/k8saudit-gke/rules/k8s_audit_rules_gke.yaml
```diff
@@ -0,0 +1,100 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2023 The Falco Authors.
```
leogr, line 3:
```suggestion
# Copyright (C) 2024 The Falco Authors.
```

leogr: Please check other file headers, too.

plugins/k8saudit-gke/rules/k8s_audit_rules_gke.yaml
```diff
@@ -0,0 +1,100 @@
...
+- list: falco_privileged_images
+  items: [
+    gke.gcr.io/netd,
+    gke.gcr.io/csi-node-driver-registrar,
+    gke.gcr.io/gcp-compute-persistent-disk-csi-driver,
+    gke.gcr.io/kube-proxy-amd64,
+    gke.gcr.io/k8s-dns-node-cache
+  ]
+  override:
+    items: append
```
leogr, line 48: > Fri Mar  8 15:32:53 2024: Loading plugin 'k8saudit-gke' from file /usr/share/falco/plugins/libk8saudit-gke.so
> Fri Mar  8 15:32:53 2024: Loading plugin 'json' from file /usr/share/falco/plugins/libjson.so
> Fri Mar  8 15:32:53 2024: Validating rules file(s):
> Fri Mar  8 15:32:53 2024:    /__w/plugins/plugins/plugins/k8saudit-gke/rules/k8s_audit_rules.yaml
> Fri Mar  8 15:32:53 2024:    /__w/plugins/plugins/plugins/k8saudit-gke/rules/k8s_audit_rules_gke.yaml
> Error: /__w/plugins/plugins/plugins/k8saudit-gke/rules/k8s_audit_rules.yaml: Ok
> /__w/plugins/plugins/plugins/k8saudit-gke/rules/k8s_audit_rules_gke.yaml: Invalid
> 1 Errors:
> In rules content: (/__w/plugins/plugins/plugins/k8saudit-gke/rules/k8s_audit_rules_gke.yaml:0:0)
>     list 'falco_privileged_images': (/__w/plugins/plugins/plugins/k8saudit-gke/rules/k8s_audit_rules_gke.yaml:38:2)
> ------
> - list: falco_privileged_images
>   ^
> ------
> LOAD_ERR_VALIDATE (Error validating rule/macro/list/exception objects): List uses 'append' or 'override.items: append' but no list by that name already exists

The `falco_privileged_images` does not exist in `k8s_audit_rules.yaml` even it is referenced (i was not noticed because in this condition `ka.req.pod.containers.image.repository in (falco_privileged_images)` it is being parsed as a bare string). cc @LucaGuerra @Issif @jasondellaluce @alacuku 

@sboschman can you add an empty `falco_privileged_images` list in `k8s_audit_rules.yaml` ? This should fix the problem.

sboschman: the list is defined in the falco stable rules, which is bundled by default... (well except in ci it seems), so not sure adding an empty list in `k8s_audit_rules.yaml` is the correct way to go. Perhaps trying to add the stable rules to ci is more in line with what happens when users are trying to activate the k8saudit or k8saudit-gke plugins

sboschman:
> The error simply means that the plugin is not present in the container image.
> I can't recall the full workflow right now. @LucaGuerra @alacuku, any clue? 🤔

@leogr , indeed, not sure what the expected behaviour is here.

1) Without the `k8saudit` rules, validation fails; sort of expected as the rules append/override the `k8saudit` rules and are not stansalone rules by themselves
2) As @Issif suggested, we can add a symlink to include the `k8saudit` rules, now we can pass the `validate-local` job, but `validate-falcoctl` job will fail as it now messes up rules and deps; it loads the deps from the `k8saudit` rules into the container, and tries to validate the `k8saudit` rules with the `k8saudit-gke.so` plugin, which it did not load into the container... and it never gets to the point to validate the `k8saudit-gke` rules.

leogr:
Hey @sboschman 

I just wanted to let you know that @LucaGuerra, @alacuku and I are working to find a solution to the two distinct problems affecting this PR:

1. There's (at least) one missing list definition in `k8s_audit_rules.yaml` (the list is `falco_privileged_images`, the one you tried to extend. @LucaGuerra will open a separate PR to address this. Once this is fixed, we will ask you to rebase this PR (and the symlink idea will likely be okay).
2. The [missing plugin artifact](https://github.com/falcosecurity/plugins/pull/424#issuecomment-1985995091) (as you reported). We are currently evaluating different options. We also were in the process of reviving our CI pipeline (which accumulated some technical debt).

We need a bit of time to address these, then we will get back to you :) 

Btw, thanks again for this significant contribution. I look forward to seeing this merged :clap:  

:pray:

sboschman:
Thanks for the heads up, @leogr . All the help is much appreciated 🙏

alacuku:
Hi @sboschman, could you please rebase your PR on top of #432? I think you should drop all the changes you have made to the CI for debugging purposes. 

The #432 unifies how the deps are extracted from the rulesfiles and handles the case when the tested plugin is set as an alternative. Furthermore, the `validate-falcoctl` job loads all the rules at once, which is needed for the k8saudit-gke plugin.

plugins/k8saudit-gke/README.md
```diff
@@ -0,0 +1,120 @@
...
+This plugin tries to reconstruct an audit event object as defined in the `audit.k8s.io` API group from the available information in the Google [`LogEntry`](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry). Unfortunately this might mean that certain information is simply unavailable to Falco.
+
+## How it works
+
+GKE Admin Activity audit logs and GKE Data Access audit logs can be send to a Pub/Sub topic using a logs routing sink. The Falco `k8saudit-gke` plugin uses a subscription on this Pub/Sub topic to pull the audit log entries.
+
+Optionally, the `k8saudit-gke` plugin can use the Google Container API to fetch cluster resource metadata labels. These cluster labels are appended to the resource labels of the log entry.
+
+Finally, the Google audit log entries are converted to a Kubernetes audit event object and handed off to the Falco rule pipeline. This means the field extraction methods and rules of the [`k8saudit`](https://github.com/falcosecurity/plugins/tree/master/plugins/k8saudit) can be used.
+Note: As the Kubernetes audit event is reconstructed from a Google audit logs entry some Falco rules might not work as expected due to missing information.
```
Issif, line 18:
```suggestion
> [!WARNING] 
> As the Kubernetes audit event is reconstructed from a Google audit logs entry some Falco rules might not work as expected due to missing information.
```

plugins/k8saudit-gke/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Configuration
+
+Here's an example of configuration of `falco.yaml`:
+
+```yaml
+plugins:
+- init_config:
+    project_id: "your-gcp-project-id"
+    maxEventSize: 8000000
```
Issif, line 38:
```suggestion
    max_event_size: 8000000
```

plugins/k8saudit-gke/hack/falco.yaml
```diff
@@ -0,0 +1,77 @@
...
+output_timeout: 2000
+plugins:
+- init_config:
+    project_id: "te000-589aa8c5b4314ba934a6"
+    fetch_cluster_metadata: false
+    max_event_size: 8000000
+  library_path: libk8saudit-gke.so
+  name: k8saudit-gke
+  # open_params: "file://test/pods_create.json"
+  open_params: "falco-gke-audit-sub"
```
Issif, line 54: are the `project_id` and `open_params` not sensitive in that example?

plugins/k8saudit-gke/hack/falco.yaml
```diff
@@ -0,0 +1,77 @@
...
+- init_config: ""
+  library_path: libjson.so
+  name: json
+priority: debug
+program_output:
+  enabled: false
+rules_file:
+- /etc/falco/falco_rules.yaml
```
Issif, line 62: As the syscall collection is disabled, we can remove it, same for all settings concerning `modern_pbf`

Issif (changes requested):
Impressive work, see my comments for some suggestions. I don't know the GCP API so I trust you for large parts.

plugins/k8saudit-gke/rules/k8s_audit_rules_gke.yaml
sboschman: updated the copyright text; not sure if github is able to track changes nicely as I am force pushing to rebase on top of the ci fixes branch

plugins/k8saudit-gke/hack/falco.yaml
sboschman: `open_params` is the PubSub subscription id (you can define your own id/name when creating the subscription. It is the same id/name used in the readme. Without a (service)account with permissions to the subscription you can't access it.

`project_id` requires Google IAM, so it is not sensitive in that sense. But yeah, no point in having my google project id in it, so I will change this to `your-gcp-project-id`, same as in the readme 👍

plugins/k8saudit-gke/hack/falco.yaml
sboschman: `falco_rules.yaml` provides at least one list, on which the base `k8saudit` rules depend. In our setup I am also including the sandbox rules as there also seems to be a rule dependency on it from base `k8saudit`, can't remember the exact dependency, guess we will find out once ci rule validations runs.

sboschman:
> > Hi @sboschman, could you please rebase your PR on top of #432? I think you should drop all the changes you have made to the CI for debugging purposes.
> > The #432 unifies how the deps are extracted from the rulesfiles and handles the case when the tested plugin is set as an alternative. Furthermore, the `validate-falcoctl` job loads all the rules at once, which is needed for the k8saudit-gke plugin.
> 
> The validation jobs are now working as intended. Failing because the rules seem not to be valid.

yeah wrong engine version, but we haven't really validated any rules yet, so I am still expecting it to fail b/c the deps on stable and/or sandbox rules.

alacuku:
> yeah wrong engine version, but we haven't really validated any rules yet, so I am still expecting it to fail b/c the deps on stable and/or sandbox rules.

This #431 PR should solve the deps issue with stable/sandbox rules. @LucaGuerra

sboschman:
> > yeah wrong engine version, but we haven't really validated any rules yet, so I am still expecting it to fail b/c the deps on stable and/or sandbox rules.
> 
> This #431 PR should solve the deps issue with stable/sandbox rules. @LucaGuerra

I see... but I rebased on #432, so that change isn't used yet... should I cherrypick the change from #431 and update the gke rules to use this new list?

plugins/k8saudit-gke/hack/falco.yaml
sboschman: rule deps are resolved with #431 , removing the falco_rules.yaml from the list

sboschman:
we do have a [green build](https://github.com/falcosecurity/plugins/actions/runs/8266143735), @leogr @alacuku @Issif  🥳

plugins/k8saudit-gke/rules/k8s_audit_rules_gke.yaml
sboschman: fixed with #431

alacuku:
#432 merged.

leogr (approved):
LGTM,

https://github.com/falcosecurity/plugins/pull/431 will followup shortly, cc @LucaGuerra

PR #425, opened 2024-03-07
feat(cloudtrail): support accounts for org trails

### LINKED ISSUE #423: Cloudtrail: Support for multiple accounts
**Motivation**

I'm using Logray with falcodump [1] and cloudtrail plugin.

For organization trails Cloudtrail uses a S3 path as follows:
`bucket_name/prefix_name/AWSLogs/O-ID/Account ID/CloudTrail/Region/YYYY/MM/DD/file_name.json.gz`. [2]

Sometimes it is necessary to search for events in several accounts at the same time.

**Feature**

Introduce support for accounts with an additional parameter `S3AccountList`. This parameter is a comma separated list of account IDs.

Application logic should be:

* Open path is `s3://bucket_name/prefix_name/AWSLog/O-ID/Account ID/` => same as before
* Open path is `s3://bucket_name/prefix_name/AWSLogs/O-ID/`
  * If `S3AccountList` is set => iterate over all accounts in this list, iterate over all regions and open according to `S3Interval` parameter
  * If `S3AccountList` is not set => get all account ID's in the bucket (ListObjects(Paginator) call with delimiter "/"), iterate over this list, iterate over all regions and open files according to `S3Interval`

Known limitation:

When `S3AccountList` and `S3Interval` are not set, getting all events over all accounts, regions and the entire period could take quite some time or even run into time outs (depending on the size of the organization).

**Alternatives**

Getting events per account manually and merge same later with `mergecap`.

**Additional context**

If there are no objections to this request, I will prepare a commit for it.

Related PR #422 

[1]: https://www.wireshark.org/docs/man-pages/falcodump.html
[2]: https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-find-log-files.html

uhei:
**What type of PR is this?**

/kind feature

/area plugins

Support accounts for organization trails. When open parameter is 's3://bucket_name/prefix_name/AWSLogs/O-ID/' and 'S3AccountList' is not set get all events for all accounts in the organization CloudTrail bucket.

If 'S3AccountList' is configured get events for the account IDs specified in the comma separated list.

S3Interval also works for organisation trails.

Fixes #423


plugins/cloudtrail/pkg/cloudtrail/source.go
```diff
@@ -200,32 +207,83 @@ func (oCtx *PluginInstance) openS3(input string) error {
...
 			intervalPrefix += "/"
 		}
 		intervalPrefix += "CloudTrail/"
-	}
+		intervalPrefixList = append(intervalPrefixList, intervalPrefix)
+	} else if awsLogsOrgRE.MatchString(prefix) {
+		if (! strings.HasSuffix(intervalPrefix, "/")) {
+			intervalPrefix += "/"
+		}
+		if len(s3AccountList) > 0 {
```
Issif, line 222:
```suggestion
		if s3AccountList != "" {
```

plugins/cloudtrail/pkg/cloudtrail/source.go
```diff
@@ -200,32 +207,83 @@ func (oCtx *PluginInstance) openS3(input string) error {
...
+			// build intervalPrefixList by using the provided S3AccountList
+			accountListArray := strings.Split(s3AccountList , ",")
```
Issif, line 224: Maybe then add a check than the length of the array > 0

Issif (changes requested):

Issif (commented):
Really good, I've just some minor suggestions. To complete the changes, could you also update the version and the changelog please

plugins/cloudtrail/pkg/cloudtrail/source.go
uhei: fixed

plugins/cloudtrail/pkg/cloudtrail/source.go
uhei: fixed

uhei:
> Really good, I've just some minor suggestions. To complete the changes, could you also update the version and the changelog please

Thanks. I've changed the version.
Generating the changelog seems broken for cloudtrail. Should be fixed in an own commit.

plugins/cloudtrail/pkg/cloudtrail/cloudtrail.go
```diff
@@ -48,7 +48,7 @@ const (
 	PluginName               = "cloudtrail"
 	PluginDescription        = "reads cloudtrail JSON data saved to file in the directory specified in the settings"
 	PluginContact            = "github.com/falcosecurity/plugins/"
-	PluginVersion            = "0.11.0"
+	PluginVersion            = "0.12.0"
```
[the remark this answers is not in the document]
uhei: A commit is still in the preparation stage. I've added concurrency to list the keys. This speeds up the whole process if all accounts in an organisation are affected (~5min instead of ~35min for approx. 900 accounts with a time interval of 1h).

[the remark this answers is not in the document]
leogr: Great!

Generally speaking, we prefer to introduce all these improvements in bulk into the 
Please let us know your plans. If you already have a list of things you want to add, I recommend opening a GitHub issue and tracking them all. It would be easier for us :pray:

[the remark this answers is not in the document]
uhei: To be honest: I don't really have a plan. I'm just working on getting my use case (analyzing organization CloudTrail with tool X) to work smoothly. #429 should be the last commit.

The changes I still need concern another project/repo.

PR #429, opened 2024-03-12
refactor(cloudtrail): Get S3 keys concurrently

uhei:
**What type of PR is this?**

/kind cleanup

/area plugins

To speed up the process of getting all the keys, divide the inputParams array into chunks and get the keys for each item in the chunk concurrently.

This should be my last commit for the time being. This means that version 0.12 can be released (after testing by third parties).

### DIFF
--- a/plugins/cloudtrail/pkg/cloudtrail/source.go
+++ b/plugins/cloudtrail/pkg/cloudtrail/source.go
@@ -53,6 +53,12 @@ const (
 	sqsMode
 )
 
+type listOrigin struct {
+	prefix *string
+	startAfter *string
+}
+
+
 type fileInfo struct {
 	name         string
 	isCompressed bool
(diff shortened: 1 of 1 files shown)


leogr:
cc @LucaGuerra @Issif

plugins/cloudtrail/pkg/cloudtrail/source.go
```diff
@@ -153,6 +161,70 @@ func (p *PluginInstance) initS3() error {
 	return nil
 }
 
+func chunkListOrigin(orgList []listOrigin, chunkSize int) [][]listOrigin {
+	if len(orgList) == 0 {
+		return nil
+	}
```
Issif, line 167: should we not also check if `chunckSize >= 1` ?

Issif (approved):

leogr (approved):

PR #431, opened 2024-03-13
cleanup(rules): remove dangling references to lists, lint

LucaGuerra:
**What type of PR is this?**

/kind cleanup

/area plugins

Some rules in the k8s_audit set were referencing lists present in `falco_rules` and `falco_sandbox_rules` sets. I believe it is risky to keep them like this because we introduce a dependency on something that is optional. While it's true that if you have both rulesfiles loaded you probably want those lists to be the same, we cannot assume that you will always have `falco_rules` loaded if you are using k8s audit and we need to leave this choice to the user.

Fixes #


leogr (commented):
There are other lists prefixed with `falco_`. I think it's worth renaming them all for consistency.

Also, I suggest `k8saudit_` (instead of just `k8s_`).

Otherwise, SGMT!

LucaGuerra:
@leogr sure, updated!

leogr (approved):

PR #461, opened 2024-04-09
update(registry.yaml): register journald plugin

gnosek:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area plugins

/area registry

> /area build

> /area documentation

This PR reserves plugin ID 514 for the journald plugin: https://github.com/gnosek/falco-journald-plugin

Fixes #

514 is the syslog port, it felt appropriate :)

### DIFF
--- a/registry.yaml
+++ b/registry.yaml
@@ -432,3 +432,22 @@ plugins:
         source: k8s_audit
       extraction:
         supported: true
+  - name: journald
+    description: Read Journald events into Falco
+    authors: Grzegorz Nosek
+    contact: https://github.com/gnosek/falco-journald-plugin
+    maintainers:
+      - name: Grzegorz Nosek
+    url: https://github.com/gnosek/falco-journald-plugin
+    license: Apache-2.0
+    keywords:
+      - syslog
+      - systemd
+      - journald
+    capabilities:
+      sourcing:
+        supported: true
+        id: 17
+        source: journal
+      extraction:
+        supported: true


registry.yaml
```diff
@@ -432,3 +432,22 @@ plugins:
         source: k8s_audit
       extraction:
         supported: true
+  - name: falco_journald_plugin
```
leogr, line 435:
```suggestion
  - name: journald
```

https://github.com/falcosecurity/plugins?tab=readme-ov-file#registering-a-new-plugin

> The plugin name must match this [regular expression](https://en.wikipedia.org/wiki/Regular_expression): ^[a-z]+[a-z0-9-_\-]*$ (however, its **not recommended** to use _ in the name, unless you are trying to match the name of a source or for particular reasons)

Furthermore, why do we need to use `falco` prefix and `plugin` suffix? Could we avoid them?

registry.yaml
```diff
@@ -432,3 +432,22 @@ plugins:
...
+    url: https://github.com/gnosek/falco-journald-plugin
+    license: Apache-2.0
+    keywords:
+      - syslog
+      - systemd
+      - journald
+    capabilities:
+      sourcing:
+        supported: true
+        id: 514
```
leogr, line 450:
```suggestion
        id: 17
```

Using the port number was nice, but unfortunately :sweat:  :point_down: 

https://github.com/falcosecurity/plugins/blob/main/docs/plugin-ids.md#assigning-an-id
>When making your request, please choose the next available ID in the [registry.yaml](https://github.com/falcosecurity/plugins/blob/main/registry.yaml) file. The id will be definitively assigned to your plugin once the corresponding PR is merged, and the [registry.yaml](https://github.com/falcosecurity/plugins/blob/main/registry.yaml) file is updated.

leogr (commented):
Otherwise, SGTM!

registry.yaml
leogr: Moreover, it has to match with the actual plugin name returned by the plugin API, which is `journald`, isn't it? :thinking: 
https://github.com/gnosek/falco-journald-plugin/blob/16137d0d9f947248dbc61acf9e4c28e0b7f494e4/src/lib.rs#L34

gnosek: Yeah. Not sure why I thought I left the crate name as the plugin name. Will update.

registry.yaml
gnosek: :sadpanda: will update ;)

leogr (approved):
:+1:

PR #463, opened 2024-04-10
feat(gcpaudit): Add Display attribute for fields

uhei:
**What type of PR is this?**

/kind feature

/area plugins

The Falco Bridge plugin of another tool is using the 'Display' value to have a name for the field. => Adding 'Display' to all fields so that this plugin is usable.


leogr:
Hey @uhei thanks!

Could you please bump the plugin version? :pray:

plugins/gcpaudit/rules/gcp_auditlog_rules.yaml
```diff
@@ -20,7 +20,7 @@
 
 - required_plugin_versions:
   - name: gcpaudit
-    version: 0.3.0
+    version: 0.3.1
```
leogr, line 23:
```suggestion
    version: 0.3.0
```

No need to bump this, since the ruleset is still compatible with plugin version 0.3.0

leogr (approved):

PR #464, opened 2024-04-11
update(CI): change release tag format

### LINKED ISSUE #223: change plugins git tag convention to be go-compatible
**Motivation**

Currently, every plugin release is marked by a git tag with the naming convention `<plugin_name>-x.y.z`. This is also used by our CI to build and release plugin artifacts.

However, many/most plugins are developed in Go and the git tags for their release cannot be used to import the plugins' code in other Go projects because they don't respect the Go version convention for modules in subdirectories (see: https://go.dev/ref/mod#vcs-version).

**Feature**

Changing the naming convention from `<plugin_name>-x.y.z` to `plugins/<plugin_name>/vx.y.z`, which is compatible with the Go package manager and improves the code reusability of each plugin. We can easily adapt our CI to use this format for building release artifacts. Two birds with one stone.

**Alternatives**

**Additional context**

alacuku:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area plugins

> /area registry

 /area build

> /area documentation

See #223 

Fixes #223

@falcosecurity/plugins-maintainers Please note that once this PR gets merged, you will need to use the new tag format to release a plugin!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,14 +18,13 @@ jobs:
       package: ${{ steps.regex-match.outputs.group1 }}
       
     steps:
-      - name: Extract semver ℹ️
+      - name: Validate tag ℹ️
         uses: actions-ecosystem/action-regex-match@v2
-        id: regex-match
         with:
           text: ${{ github.ref_name }}
-          regex: '^([a-z]+[a-z0-9_\-]*)-(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-((0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*))?$'
-          
-      - name: Fail on non semver tag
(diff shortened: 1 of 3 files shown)


leogr (commented):
Could we update the doc :point_down: as well, please?
https://github.com/falcosecurity/plugins/blob/main/release.md#tag-format-stable-build-only

Otherwise, SGTM! :clap:

.github/workflows/release.yml
```diff
@@ -18,14 +18,13 @@ jobs:
       package: ${{ steps.regex-match.outputs.group1 }}
       
     steps:
-      - name: Extract semver ℹ️
+      - name: Validate tag ℹ️
```
leogr, line 21: What does it mean `i` ? :thinking:

leogr (commented):
Noticed that the `pre` case must also be fixed to accommodate the possibility of new and old tags, see :point_down: 

https://github.com/falcosecurity/plugins/pull/464/files#diff-ea759ade72ddb523d1113744fdbc8c8ab7bc9c5cc60da3a854a78d361ca430bcL166

.github/workflows/release.yml
alacuku: It was already there.

leogr: I guess it was a leftover. I recommend to clean it up, eventually

leogr (approved):

Issue #148, opened 2022-09-15
commit and release note naming conventions


jasondellaluce:
**What to document**

We have no definite naming conventions for either commits or release notes (I don't think we have release notes at all).

In Falco, we have a naming convention for release notes touching rulesets (see: https://github.com/falcosecurity/.github/blob/main/CONTRIBUTING.md#rule-type), but although this repository is the official place for all our plugin rulesets, we haven't something like this here.

At the same time, our changelog generation tool looks for commit messages following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) specifications. For a given plugin, the changelog will filter commit messages in the form of `xxx(plugins): ...` or `xxx(plugins/<plugin_name>): ...`. However, there is nothing enforcing this at the CI level, not even with a warning, which ends up potentially missing valuable commits in the changelogs.

**Proposals/Idea*
- Add a CI job checking that:
  - If a file changes in a `plugin/<plugin_name>` path, then the commit must follow the `xxx(plugins/<plugin_name>): ...` message pattern
  - If a file changes in more than one `plugin/<plugin_name>` path, then the commit must follow the `xxx(plugins): ...` message pattern
- Check or enforce than the `rule(xxx):...` commit message format is respected, even in a looser form than what we have in Falco

PR #485, opened 2024-05-03
Updat changelog scripts to support the recently changed plugin tag convention

sboschman:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area plugins

> /area registry

/area build

> /area documentation

Support changelog generation with the changed plugin tags (https://github.com/falcosecurity/plugins/issues/223)

Fixes #

### DIFF
--- a/build/changelog/changelog.go
+++ b/build/changelog/changelog.go
@@ -130,7 +130,7 @@ func main() {
 		fail(err)
 	}
 
-	var rgx, rgxAlt *regexp.Regexp
+	var rgx, rgxAlt, rgxDeps *regexp.Regexp
 	if len(plugin) > 0 {
 		// craft a regex to filter all plugin-related commits that follow
 		// the conventional commit format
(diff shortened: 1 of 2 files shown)


build/changelog/changelog.go
```diff
@@ -127,15 +130,26 @@ func main() {
...
 	if len(plugin) > 0 {
 		// craft a regex to filter all plugin-related commits that follow
 		// the conventional commit format
 		rgx, _ = regexp.Compile("^[a-f0-9]+ [a-zA-Z]+\\(([a-zA-Z\\/]+\\/)?" + plugin + "(\\/[a-zA-Z\\/]+)?\\):.*")
+
+		// alternative plugin name used as scope
+		if plugin == "k8saudit" {
+			pluginAlt := "k8s_audit"
+			rgxAlt, _ = regexp.Compile("^[a-f0-9]+ [a-zA-Z]+\\(([a-zA-Z\\/]+\\/)?" + pluginAlt + "(\\/[a-zA-Z\\/]+)?\\):.*")
+		}
```
leogr, line 143: Why do we need this?

Is there any valid alternative that does not require hard code the plugin name? :thinking:

sboschman: There are a few commits for release 0.8 that use a different conventional commit scope (`k8s_audit` instead of `k8saudit`, e.g.:

```
v0.8
cleanup(plugins/k8s_audit): make the rulesefile self-referenced

v0.6
feat(plugins/k8saudit): include query params in health check endpoint exceptions
```

Pretty much summed up by https://github.com/falcosecurity/plugins/issues/148

leogr: A possible solution for the general case: 
since [k8s_audit](https://github.com/falcosecurity/plugins/blob/0e7ef76987476224a799e6db8af57b5716040b30/registry.yaml#L58) is the _source_ name (which may differ from the _plugin_ name), we may use the _source_ as an alternative name used as scope.

If we want to postpone this fix for a follow-up PR, I'm okay with that. In such a case, please leave a comment in the code (e.g., `fixme: ....`) explaining why we needed that workaround.

sboschman: checking the commit scope if it contains the event source name now

leogr (approved):

PR #490, opened 2024-05-06
plugins(k8saudit-gke): export ConvertLogEntry method

SKosier:
**What type of PR is this?**

/kind feature

/area plugins

For the project I have we are ingesting gke events in other way and we would like to leverage already existing OSS code for converting gke events to `k8s_audit` ones, due to maintainability reasons.

Fixes #

### DIFF
--- a/plugins/k8saudit-gke/pkg/k8sauditgke/convert.go
+++ b/plugins/k8saudit-gke/pkg/k8sauditgke/convert.go
@@ -31,6 +31,18 @@ import (
 	auditv1 "k8s.io/apiserver/pkg/apis/audit/v1"
 )
 
+// ConvertLogEntry converts a logging.LogEntry and an audit.AuditLog into an auditv1.Event.
+//
+// Parameters:
+//   - logEntry: A pointer to a logging.LogEntry representing the log entry to convert.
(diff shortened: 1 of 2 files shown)


jasondellaluce:
@sboschman @leogr what's your take here? I think the conversion method genuinely offers a valuable piece of functionality useful to other projects and that perhaps we could export as part of the plugin's package. LGTM.

sboschman:
I don't mind exposing the conversion method as part of the plugin sdk. Seems a bit strange to just expose a random internal function b/c it might be useful without any method signature guarantee or being documented as part of the sdk.

As sdk changes are a bit of work, the method can be made public for the time being. Just add a comment indicating that it is an internal method without any signature guarantees. So use at your own risk.

SKosier:
> I don't mind exposing the conversion method as part of the plugin sdk. Seems a bit strange to just expose a random internal function b/c it might be useful without any method signature guarantee or being documented as part of the sdk.
> 
> As sdk changes are a bit of work, the method can be made public for the time being. Just add a comment indicating that it is an internal method without any signature guarantees. So use at your own risk.

Thank you @sboschman  for the review. I applied the feedback and added the comment above the method signature.

sboschman (approved):

PR #506, opened 2024-06-09
update(cloudtrail): Update ct.resources handling

geraldcombs:
Generalize our array length check. Use strings.TrimSuffix in order to avoid a potential string under-read.

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area plugins

> /area registry

> /area build

> /area documentation

Fixes #

### DIFF
--- a/plugins/cloudtrail/pkg/cloudtrail/extract.go
+++ b/plugins/cloudtrail/pkg/cloudtrail/extract.go
@@ -532,16 +532,14 @@ func getfieldStr(jdata *fastjson.Value, field string) (bool, string) {
 	case "ct.resources":
 		var resources string = ""
 		rlist := jdata.GetArray("resources")
-		if rlist == nil {
+		if rlist == nil || len(rlist) == 0 {
 			return false, ""
 		}
 		for _, resource := range rlist {
 			resources += string(resource.MarshalTo(nil))
 			resources += ","
 		}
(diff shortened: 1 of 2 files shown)


plugins/cloudtrail/pkg/cloudtrail/extract.go
```diff
@@ -532,16 +532,14 @@ func getfieldStr(jdata *fastjson.Value, field string) (bool, string) {
 	case "ct.resources":
 		var resources string = ""
 		rlist := jdata.GetArray("resources")
-		if rlist == nil {
+		if len(rlist) == 0 {
```
leogr, line 535:
```suggestion
		if rlist == nil || len(rlist) == 0 {
```
As per `GetArray` documentation, it can still return `nil`:
> nil is returned for non-existing keys path or for invalid value type.

leogr (commented):
Hey @geraldcombs 
Just a minor thing, otherwise SGTM!

Also, can you bump the plugin version to `0.12.3`, please? :pray: 

Thanks

plugins/cloudtrail/pkg/cloudtrail/extract.go
geraldcombs: Done, along with the version change.

leogr (approved):

PR #527, opened 2024-09-11
feat: Added keycloak plugin to registry.yaml

mattiaforc:
**What type of PR is this?**

/kind feature

/area registry

Introduced a new Falco plugin with source and extractor capabilities for Keycloak user/admin events.

Fixes #

### DIFF
--- a/registry.yaml
+++ b/registry.yaml
@@ -498,3 +498,22 @@ plugins:
         source: gitlab
       extraction:
         supported: true
+  - name: keycloak
+    description: Falco plugin for sourcing and extracting Keycloak user/admin events
+    authors: Mattia Forcellese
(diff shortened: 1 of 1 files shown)


Issif (commented):
Your PR and your plugin are OK for me (thanks and congrats), but I don't see any OCI packages related to your repo, even if the workflow is correct. Do you know why?

mattiaforc:
Oh, it looks like that packages that were created while the repository was private remain private even if the repository's visibility changes. I set it to public now and I managed to pull it:

```bash
docker pull ghcr.io/mattiaforc/falco-keycloak-plugin/plugin/keycloak:0.1 
```

Thanks for the swift reply and feedback!

leogr (approved):

PR #528, opened 2024-09-12
fix:Do not output information that contains confidential data.

pyama86:
**What type of PR is this?**
/kind bug

/area plugins

This rule has an issue where it detects a configmap that seems to contain confidential information, but still outputs its content. For example, the content could potentially be notified to Slack. Therefore, it would be better not to output the object's content in this case.

Fixes #


leogr (approved):

alacuku (commented):
Thanks @pyama86!
/lgtm

pyama86:
Thank you for merging. I always appreciate your wonderful contributions.

PR #537, opened 2024-11-06
fix(proc/scan): scan all existing threads

alacuku:
By scanning only /proc we missed the threads of a given process. It scans all threads under a given process in /proc/pid/task.

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

 /area plugins

> /area registry

> /area build

> /area documentation

By scanning only /proc we missed the threads of a given process.
It scans all threads under a given process in /proc/pid/task.

Fixes #

### DIFF
--- a/plugins/k8smeta/src/plugin.cpp
+++ b/plugins/k8smeta/src/plugin.cpp
@@ -302,42 +302,65 @@ void my_plugin::do_initial_proc_scan()
             continue;
         }
 
-        // Example of the path: `/proc/1/cgroup`
-        proc_path = std::string(proc_root)
-                            .append("/")
-                            .append(file_name.c_str())
-                            .append("/cgroup");
+        // Now scan /proc/pid/task for threads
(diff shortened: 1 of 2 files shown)


plugins/k8smeta/src/plugin.cpp
```diff
@@ -302,42 +302,65 @@ void my_plugin::do_initial_proc_scan()
             continue;
         }
 
-        // Example of the path: `/proc/1/cgroup`
-        proc_path = std::string(proc_root)
-                            .append("/")
-                            .append(file_name.c_str())
-                            .append("/cgroup");
+        // Now scan /proc/pid/task for threads
+        std::string task_dir = proc_root + "/" + file_name.c_str() + "/task";
```
LucaGuerra, line 306: `std::string` + `char*` + `char*` + `char*` should result in an error, why not `file_name` instead of `file_name.c_str()`?

LucaGuerra (approved):

PR #544, opened 2024-11-20
Expose gcp.resourceLabels field in the gcp plugin

pirrofra:
**What type of PR is this?**
/kind feature

/area plugins

On the GCP Plugin we already expose the resource type via the `gcp.resourceType` field. Given that the log-entry also supports a `labels` field which can contains useful information for correctly identifying a resource on GCP, we want to expose that using a new `gcp.resourceLabels` field.
This field will contain the JSON representation of the labels attached on the [cloud log entry.](https://cloud.google.com/logging/docs/reference/v2/rest/v2/MonitoredResource)

Fixes #

### DIFF
--- a/plugins/gcpaudit/pkg/gcpaudit/gcpaudit.go
+++ b/plugins/gcpaudit/pkg/gcpaudit/gcpaudit.go
@@ -30,7 +30,7 @@ const (
 	PluginName               = "gcpaudit"
 	PluginDescription        = "Read GCP Audit Logs"
 	PluginContact            = "github.com/falcosecurity/plugins"
-	PluginVersion            = "0.4.0"
+	PluginVersion            = "0.5.0"
 	PluginEventSource        = "gcp_auditlog"
 )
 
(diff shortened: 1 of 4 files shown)


SKosier:
Two small comments:

Please bump the version of the plugin `[here](https://github.com/falcosecurity/plugins/blob/main/plugins/gcpaudit/pkg/gcpaudit/gcpaudit.go#L33)`
And add the new extraction to the README

plugins/gcpaudit/pkg/gcpaudit/gcpaudit.go
```diff
@@ -30,7 +30,7 @@ const (
 	PluginName               = "gcpaudit"
 	PluginDescription        = "Read GCP Audit Logs"
 	PluginContact            = "github.com/falcosecurity/plugins"
-	PluginVersion            = "0.4.0"
+	PluginVersion            = "0.4.1"
```
LucaGuerra, line 33:
```suggestion
	PluginVersion            = "0.5.0"
```

PR #545, opened 2024-11-20
Expose gcp.resourceLabels field in the gcp plugin

pirrofra:
**What type of PR is this?**
/kind feature

/area plugins

On the GCP Plugin we already expose the resource type via the `gcp.resourceType` field. Given that the log-entry also supports a `labels` field which can contains useful information for correctly identifying a resource on GCP, we want to expose that using a new `gcp.resourceLabels` field.
This field will contain the JSON representation of the labels attached on the [cloud log entry.](https://cloud.google.com/logging/docs/reference/v2/rest/v2/MonitoredResource)

Fixes #


LucaGuerra (approved):
Thank you! LGTM

PR #548, opened 2024-12-09
new(plugins/k8saudit): add subject name fields

LucaGuerra:
**What type of PR is this?**

/kind feature

/area plugins

Sometimes people need to match/display k8s audit events against the actual user that is in the subject section. These fields will split them between regular users, groups and service accounts.

Fixes #


leogr:
SMGT to me, I will give it a second look before the final approve.

@LucaGuerra @Issif 
Note for ourselves: we will need to update dependent plugins as well (i.e., aks, -eks, -like...)

PR #625, opened 2025-02-13
new(plugins/k8smeta): update k8smeta plugin to require plugin API version 3.9.0.

FedeDP:
**What type of PR is this?**

/kind cleanup
/kind feature

/area plugins

Implemented the "suggested output fields feature", by suggesting `k8smeta.pod.name` and `k8smeta.ns.name` as output fields.

Also, entirely avoid the proc scan, instead relying on the listening CAPability to initially loop over the thread table to attach pod_uid to threads.
Following the above, `hostProc` initConfig key is now deprecated and unused.

Moved the plugin to 0.3.0 too (given the minimum plugin API version changes).

Fixes #

### DIFF
--- a/plugins/k8smeta/cmake/modules/plugin-sdk-cpp.cmake
+++ b/plugins/k8smeta/cmake/modules/plugin-sdk-cpp.cmake
@@ -6,7 +6,7 @@ message(
 FetchContent_Declare(
   plugin-sdk-cpp
   GIT_REPOSITORY https://github.com/falcosecurity/plugin-sdk-cpp.git
-  GIT_TAG 2097bdb5a5d77f3f38162da1f438382912465340)
+  GIT_TAG 0.2.0)
 
 FetchContent_MakeAvailable(plugin-sdk-cpp)
 set(PLUGIN_SDK_INLCUDE "${plugin-sdk-cpp_SOURCE_DIR}/include")
--- a/plugins/k8smeta/src/plugin.cpp
+++ b/plugins/k8smeta/src/plugin.cpp
@@ -256,6 +256,13 @@ void my_plugin::parse_init_config(nlohmann::json& config_json)
             }
         }
     }
+
+    // TODO: clean this up after deprecation period is over
+    if(config_json.contains(nlohmann::json::json_pointer("/hostProc")))
+    {
+        SPDLOG_WARN("hostProc init config key is deprecated since it is no "
+                    "more in use");
+    }
 }
 
 bool my_plugin::init(falcosecurity::init_input& in)
--- a/plugins/k8smeta/src/plugin.cpp
+++ b/plugins/k8smeta/src/plugin.cpp
@@ -138,8 +138,8 @@ falcosecurity::init_schema my_plugin::get_init_schema()
 		},
 		"hostProc": {
 			"type": "string",
-			"title": "Path to reach the '/proc' folder we want to scan.",
-			"description": "The plugin needs to scan the '/proc' of the host on which is running. In Falco usually we put the host '/proc' folder under '/host/proc' so the the default for this config is '/host'. The path used here must not have a final '/'."
+			"title": "[DEPRECATED] Path to reach the '/proc' folder we want to scan.",
+			"description": "[DEPRECATED] The plugin needs to scan the '/proc' of the host on which is running. In Falco usually we put the host '/proc' folder under '/host/proc' so the the default for this config is '/host'. The path used here must not have a final '/'."
 		}
 	},
 	"additionalProperties": false,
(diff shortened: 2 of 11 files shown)


plugins/k8smeta/cmake/modules/plugin-sdk-cpp.cmake
```diff
@@ -6,7 +6,7 @@ message(
 FetchContent_Declare(
   plugin-sdk-cpp
   GIT_REPOSITORY https://github.com/falcosecurity/plugin-sdk-cpp.git
-  GIT_TAG 2097bdb5a5d77f3f38162da1f438382912465340)
+  GIT_TAG 0.2.0)
```
FedeDP, line 9: Use tagged sdk-cpp version.

plugins/k8smeta/src/plugin.cpp
```diff
@@ -138,8 +138,8 @@ falcosecurity::init_schema my_plugin::get_init_schema()
 		},
 		"hostProc": {
 			"type": "string",
-			"title": "Path to reach the '/proc' folder we want to scan.",
-			"description": "The plugin needs to scan the '/proc' of the host on which is running. In Falco usually we put the host '/proc' folder under '/host/proc' so the the default for this config is '/host'. The path used here must not have a final '/'."
+			"title": "[DEPRECATED] Path to reach the '/proc' folder we want to scan.",
```
FedeDP, line 141: Deprecated initConfig key, unused now.

plugins/k8smeta/src/plugin.cpp
```diff
@@ -428,11 +295,20 @@ bool my_plugin::init(falcosecurity::init_input& in)
     try
     {
         m_thread_table = t.get_table(THREAD_TABLE_NAME, st::SS_PLUGIN_ST_INT64);
+
+        // get the 'cgroups' field accessor from the thread table
+        m_thread_field_cgroups = m_thread_table.get_field(
```
FedeDP, line 300: We need these to be able to correctly fetch cgroups string from the thread entry in the `capture_open()` listen CAP callback.

plugins/k8smeta/src/plugin.cpp
```diff
@@ -441,11 +317,53 @@ bool my_plugin::init(falcosecurity::init_input& in)
...
+// Listen capability
+//////////////////////////
+
+bool my_plugin::capture_open(const falcosecurity::capture_listen_input& in) {
+    using st = falcosecurity::state_value_type;
+
+    SPDLOG_DEBUG("enriching initial thread table entries");
+    auto& tr = in.get_table_reader();
+    auto& tw = in.get_table_writer();
+    m_thread_table.iterate_entries(
```
FedeDP, line 333: We iterate over all thread table entries, and for each, we iterate over its cgroups vector, taking the `->second` (cgroups are a vector of std::pair and `second` hosts the cgroup path, while `first` hosts the cgroup controller); for each cgroup path we try to extract the `pod_uid` and, if successful, we break the loop for the current thread entry.

plugins/k8smeta/src/plugin.cpp
```diff
@@ -520,7 +438,7 @@ std::vector<falcosecurity::field_info> my_plugin::get_fields()
     // Use an array to perform a static_assert one the size.
     const falcosecurity::field_info fields[] = {
             {ft::FTYPE_STRING, "k8smeta.pod.name", "Pod Name",
-             "Kubernetes pod name."},
+             "Kubernetes pod name.", {}, false, {}, true}, // use as suggested output format
```
FedeDP, line 441: Mark this field as suggested output format.

plugins/k8smeta/src/plugin.cpp
```diff
@@ -535,7 +453,7 @@ std::vector<falcosecurity::field_info> my_plugin::get_fields()
             {ft::FTYPE_STRING, "k8smeta.pod.ip", "Pod Ip", "Kubernetes pod ip"},
 
             {ft::FTYPE_STRING, "k8smeta.ns.name", "Namespace Name",
-             "Kubernetes namespace name."},
+             "Kubernetes namespace name.", {}, false, {}, true}, // use as suggested output format
```
FedeDP, line 456: Mark this field as suggested output format.

plugins/k8smeta/src/plugin.h
```diff
@@ -282,3 +286,4 @@ FALCOSECURITY_PLUGIN(my_plugin);
 FALCOSECURITY_PLUGIN_FIELD_EXTRACTION(my_plugin);
 FALCOSECURITY_PLUGIN_ASYNC_EVENTS(my_plugin);
 FALCOSECURITY_PLUGIN_EVENT_PARSING(my_plugin);
+FALCOSECURITY_PLUGIN_CAPTURE_LISTENING(my_plugin);
```
FedeDP, line 289: New capability.

plugins/k8smeta/src/plugin.h
```diff
@@ -245,7 +250,7 @@ class my_plugin
     private:
     // Async thread
     std::thread m_async_thread;
-    std::atomic<bool> m_async_thread_quit;
+    std::atomic<bool> m_async_thread_quit = false;
```
FedeDP, line 253: Compiler warning.

plugins/k8smeta/src/plugin.cpp
leogr: Do you confirm that this setting won't produce any effect anymore?

If so, we should add a notice in addition to the deprecation. Make sense?

leogr (commented):
Nice!

Just left two suggestions. I will do a second pass later.

Thank you

FedeDP:
@leogr pushed changes to address your review! Thanks!

plugins/k8smeta/src/plugin.cpp
FedeDP: Done

FedeDP:
Also, we might want to use the provided logger by the sdk, like i did here for the container plugin: https://github.com/FedeDP/container_plugin/pull/26

This would allow use to drop the `spdlog` dep, to have better unified logging, and to drop the `verbosity` init config key, since we would then use the verbosity used for libs logger (https://github.com/falcosecurity/libs/blob/3d1d4a930d3a16f3e72ab62876ba0837c02d2ec5/userspace/libsinsp/plugin.cpp#L79).

FedeDP:
> I will make the change in its own commit.

Won't do that here since it would be a bit too big. We can fix it later if needed.

FedeDP:
/unhold
Let's go!

PR #695, opened 2025-04-01
new(plugins/container): allow to specify which hook to be attached between {"create", "start"}

FedeDP:
**What type of PR is this?**

/kind feature

/area plugins

Some fields (namely IP address for docker engine) are not available at `Create` hook. Allow user to specify which hooks to attach.
By default, we only attach `Create` hook since we can be sure that it is called way before the container gets started, thus it gives us time to retrieve metadata before the first process starts.
If both hooks are attached, you'll get doubled "container" events, and a single "container_removed" event.

Fixes #

### DIFF
--- a/plugins/container/src/plugin_config_schema.h
+++ b/plugins/container/src/plugin_config_schema.h
@@ -27,7 +27,7 @@ const char plugin_schema_string[] = LONG_STRING_CONST(
         ]
       },
       "title": "Hooks to be attached.",
-      "description": "Hooks to be attached from the engines SDKs."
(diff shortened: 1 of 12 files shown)


plugins/container/src/plugin_config_schema.h
FedeDP, line 1: Reformatted + added new:
``
"hooks": {
      "type": "array",
      "items": {
        "enum": [
          "create",
          "start"
        ]
      },
      "title": "Hooks to be attached.",
      "description": "Hooks to be attached from the engines SDKs."
    },
```

leogr (approved):

PR #715, opened 2025-04-16
fix(plugins/container): some container plugin improvements and fixes

FedeDP:
**What type of PR is this?**

/kind bug
/kind cleanup
/kind feature

**Any specific area of the project related to this PR?**

/area plugins

**What this PR does / why we need it**:

Multiple things (split by commit):
* drop useless inotifier impl
* added a debug log with all actually connected engine sockets
* immediately enrich plugin cache with pre-existing container's metadata

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

PR #721, opened 2025-04-17
fix(plugin/container): avoid overwriting host container info when loading pre-existing containers

FedeDP:
**What type of PR is this?**

/kind bug

/area plugins

#715  loaded pre-existing containers and then copied them to the containers cache; doing so, it was overriding the `""` (empty) key, that is used for host container info.

Fixes #


alacuku (approved):
/lgtm

leogr (approved):

PR #894, opened 2025-07-24
fix(plugins/container): properly cleanup stale container cache entries for exiting containers

FedeDP:
**What type of PR is this?**

/kind bug

/area plugins

Normally, we receive "removed" events from the container runtime; but that's not the case for libvirt-lxc, lxc and bpm.
In these cases, properly cleanup container cache upon init_container process exiting.

Fixes #


ekoops (approved):
/approve

PR #897, opened 2025-07-24
fix(plugins/container): parse_exit_process_event

deepskyblue86:
**What type of PR is this?**

/kind bug

/area plugins

Apply the same logic of [libs ~user_group_updater](https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/user.h#L259)

Fixes #


ekoops (approved):
/approve

Issue #893, opened 2025-07-23
k8saudit-eks losing events


wangwillian0:
The plugin isn't fetching all events from CloudWatch, which means some events are not being evaluated against the rules.

The exact cause of the problem is at the https://github.com/falcosecurity/plugins/blob/main/shared/go/aws/cloudwatchlogs/cloudwatch.go#L125-L146.

The current implementation sets the start time using filters.SetStartTime(lastEventTime + 1). If multiple log events share the exact same millisecond timestamp, this logic will process only the first event. It then updates lastEventTime to that event's timestamp and begins the next search after that timestamp. This inherently causes any subsequent events with the same timestamp to be skipped. 

My idea to fix it temporarily would be to use `EndTime` and force a delay of a few seconds, enough for AWS to "be sure" of all the events at a specific timestamp. There is a POC of the bug below (yes, it was aided by a LLM).

`./poc -group /aws/eks/<cluster>/cluster --prefix kube-apiserver-audit` - run for a few seconds and it will show that two queries including the same timestamp can return a different number of events

`./poc -group /aws/eks/<cluster>/cluster --prefix kube-apiserver-audit --delay 10s` -  It doesn't fetch events from the past 10s at all. I haven't tested enough, but it improves the situation by a lot.

                                                                                      
```
package main

import (
	"context"
	"encoding/json"
	"flag"
	"fmt"
	"os"
	"os/signal"
	"syscall"
	"time"

	"github.com/aws/aws-sdk-go/aws"
	"github.com/aws/aws-sdk-go/aws/session"
	"github.com/aws/aws-sdk-go/service/cloudwatchlogs"
)

func main() {
	// Command-line flags
	var (
		region       = flag.String("region", "us-east-1", "AWS region")
		logGroup     = flag.String("group", "", "Log group name (required)")
		prefix       = flag.String("prefix", "", "Log stream name prefix (optional)")
		pollInterval = flag.Duration("interval", 5*time.Second, "Polling interval")
		initialShift = flag.Duration("shift", 5*time.Second, "Initial lookback shift")
		forcedDelay  = flag.Duration("delay", 0*time.Second, "Ending shift for log events")
	)
	flag.Parse()

	if *logGroup == "" {
		flag.Usage()
		os.Exit(1)
	}

	// AWS session and client
	sess := session.Must(session.NewSession(&aws.Config{Region: aws.String(*region)}))
	client := cloudwatchlogs.New(sess)

[...]

PR #899, opened 2025-07-25
fix(plugins/k8saudit-eks): Improve handling of events with the same timestamp

wangwillian0:
**What type of PR is this?**

/kind bug

/area plugins

Cloudwatch events can have the same timestamp but different IngestionTime, making the logic of polling with `filters.SetStartTime(lastEventTime + 1)` flawed against these late events.

The idea here is to query again the events from the `lastEventTime` and filter out any duplicates. This shouldn't have a relevant performance penalty since the number of events with the same timestamp is very small compared to the number of events fetched when the polling interval is in the order of seconds.

[According to the AWS docs for `FilterLogEvents`](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_FilterLogEvents.html):

> The returned log events are sorted by event timestamp, the timestamp when the event was ingested by CloudWatch Logs, and the ID of the PutLogEvents request.

Also:

> Partially full or empty pages don't necessarily mean that pagination is finished. If the results include a nextToken, there might be more log events available. You can return these additional log events by providing the nextToken in a subsequent FilterLogEvents operation. If the results don't include a nextToken, then pagination is finished. 

So I removed the page length check. `FilterLogEventsPagesWithContext` should also call the function with `lastPage` set to true, so doesn't make sense to have a separate check. 

```
- if len(page.Events) == 0 {
-     return false
- }
```

Partially fixes https://github.com/falcosecurity/plugins/issues/893

In my tests I found that events with the same timestamp, can have up to 15s of difference in the `IngestionTime`. The number is probably larger since I only checked logs for a few hours.

This doesn't solve the issue completely because it depends on how large the polling interval is set. If the polling interval is 5 seconds and an event of the same timestamp takes 10s to be ingested by AWS, it will be missed anyways.

### DIFF
--- a/shared/go/aws/cloudwatchlogs/cloudwatch.go
+++ b/shared/go/aws/cloudwatchlogs/cloudwatch.go
@@ -120,11 +120,11 @@ func (client *Client) Open(context context.Context, filter *Filter, options *Opt
 	go func() {
 		defer close(eventC)
 		defer close(errC)
+		var (
+			lastEventTime     int64
+			lastIngestionTime int64
+		)
 		for {
-			var (
-				lastEventTime     int64
-				lastIngestionTime int64
-			)
 			err := client.CloudWatchLogs.FilterLogEventsPagesWithContext(aws.Context(context), filters,
 				func(page *cloudwatchlogs.FilterLogEventsOutput, lastPage bool) bool {
 					for _, i := range page.Events {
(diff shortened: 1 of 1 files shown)


leogr:
> Thanks for your pull request. Before we can look at it, you'll need to add a 'DCO signoff' to your commits.
> 
> 📝 **Please follow instructions in the [contributing guide](https://github.com/falcosecurity/.github/blob/main/CONTRIBUTING.md) to update your commits with the DCO**
> 
> Full details of the Developer Certificate of Origin can be found at [developercertificate.org](https://developercertificate.org/).
> 
> **The list of commits missing DCO signoff**:
> 
> * [a58f83b](https://github.com/falcosecurity/plugins/commits/a58f83b43e2e07bdf1723242694bfcd27c495dd1) fix variable context

Can you please sign off your commits? :pray: 

see https://github.com/falcosecurity/.github/blob/main/CONTRIBUTING.md#developer-certificate-of-origin

wangwillian0:
Signed!

PR #901, opened 2025-07-28
fix typo from #899

wangwillian0:
There is a typo and `lastEventTime` is not updated at all. The line `lastIngestionTime = *i.IngestionTime` is duplicated. Sorry.

https://github.com/falcosecurity/plugins/pull/899


ekoops (approved):
/approve

PR #908, opened 2025-07-29
chore(plugins/container): some improvements

FedeDP:
**What type of PR is this?**

/kind cleanup

/area plugins

This PR is split into multiple commits:
* port docker engine away from deprecated APIs
* podman `system.Events` API [now](https://github.com/containers/podman/commit/768ad8653af4cebd069a865a5dbf4a7ec8517444) returns error synchronously, and it spawns its own goroutine to poll stream of events
* fixed build under recent gcc by including `algorithm` header
* improved exit strategy for goroutine workers whose stream of events gets closed
* added some tests around 'workerLoop()' API

Fixes #

/cc @nenioscio

### DIFF
--- a/plugins/container/go-worker/worker_test.go
+++ b/plugins/container/go-worker/worker_test.go
@@ -1,0 +1,132 @@
+package main
+
+import (
+	"context"
+	"github.com/falcosecurity/plugins/plugins/container/go-worker/pkg/container"
+	"github.com/falcosecurity/plugins/plugins/container/go-worker/pkg/event"
+	"github.com/stretchr/testify/assert"
+	"math"
+	"runtime"
+	"sync"
+	"testing"
+	"time"
+)
+
+type noopEngine struct {
+	exitAfter  time.Duration
+	eventAfter time.Duration
+}
+
+func (n *noopEngine) Name() string {
+	return "noop"
+}
+
+func (n *noopEngine) Sock() string {
+	panic("implement Sock")
+}
+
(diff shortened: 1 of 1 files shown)


plugins/container/go-worker/pkg/container/podman.go
```diff
@@ -271,56 +270,41 @@ func (pc *podmanEngine) Listen(ctx context.Context, wg *sync.WaitGroup) (<-chan
...
+			close(cancelChan)
+			close(outCh)
+		}()
 		size := config.GetWithSize()
 		// Blocking: convert all events from podman to json strings
 		// and send them to the main loop until the channel is closed
 		for {
 			select {
 			case <-ctx.Done():
+				cancelChan <- true
```
FedeDP, line 296: Cancel the podman managed goroutine!

plugins/container/go-worker/worker.go
```diff
@@ -48,13 +48,17 @@ func workerLoop(ctx context.Context, cb asyncCb, containerEngines []container.En
 	}
 
 	for {
-		chosen, val, _ := reflect.Select(cases)
+		chosen, val, recvOk := reflect.Select(cases)
```
FedeDP, line 51: In case we fail to recv the value, it means that the channel has been closed, and thus the worker goroutine at idx `chosen` has left. Drop it from our cases slice.

FedeDP:
@nenioscio also added some tests around `workerLoop` behavior :)

PR #911, opened 2025-08-01
new(plugins/container): properly send `container_removed` events for bpm,lxc,libvirt_lxc too

FedeDP:
**What type of PR is this?**

/kind feature

/area plugins

Let the async event cleanup the cache for us.
This way, other plugins can get notified even when we are removing a bpm/lxc container.

Also, do not attach the `removed` SDK listener from the plugin, since we will only rely upon `procexit` event
to trace exited containers and cleanup plugin cache.

Also, fix a subtle issue: allow to extract metadata fields from `container_removed` events too, eg:
```
1971925) [11:19:32.532080721] container_removed | image: debian:11:11 | false
```

NOTE: the go-worker will still attach `removed` SDK listeners by default, for the `make exe` worker executable and for go-worker tests. But its value is always enforced by the C++ side of the plugin when calling StartWorker.

Finally, fixed a segfault case triggered by possibly using `generate_async_event()` method when `m_async_ctx`  (and thus `s_async_handler`s) is null. 

Fixes #


plugins/container/src/caps/parse/parse.cpp
```diff
@@ -192,7 +195,20 @@ bool my_plugin::parse_exit_process_event(
                 m_logger.log(fmt::format("Removing container from procexit: {}",
                                          container_id),
                              falcosecurity::_internal::SS_PLUGIN_LOG_SEV_TRACE);
+#ifdef _HAS_ASYNC
```
FedeDP, line 198: Improvement on top of #894 : instead of immediately removing the container cache entry without notification, send the `container_removed` event and let our default parsing logic (the same used for events generated by the go-worker) kick in to manage the removal for us.
This way, other plugins can parse the `container_removed` events for rocket, lxc and libvirt-lxc too.

FedeDP:
I don't get why falco-tests are failing :/

plugins/container/src/caps/async/async.tpp
```diff
@@ -48,9 +48,6 @@ void generate_async_event(const char *json, bool added, bool initial_state)
     }
     else
     {
-        // set ts = now + 1s to leave some space for enqueued syscalls to be
-        // enriched
-        enc.set_ts(get_current_time_ns(1));
```
FedeDP, line 53: We don't need this trick anymore, since we will now send the `removed` event after the final `procexit` in the container.

FedeDP:
> I don't get why falco-tests are failing :/

Still don't get it :/

FedeDP:
> Still don't get it :/

Solved!

plugins/container/src/caps/parse/parse.cpp
FedeDP: > always rely upon procexit event parsing to emit container_removed event thus dropping the removed listeners on go-worker SDKs

I chose this one, that is more coherent and simple.

PR #994, opened 2025-09-23
fix(container): containerd sandbox detection

deepskyblue86:
**What type of PR is this?**

/kind bug

/area plugins

Regular containers on containerd get flagged as sandbox (`is_pod_sandbox` true), while sandbox containers get `is_pod_sandbox` false.
Moreover, in some cases `pod_sandbox_id` is empty.
This change addresses those issues by checking labels and annotations.

Fixes #


ekoops (approved):
/approve

PR #996, opened 2025-09-24
fix(container): cri sandbox detection

deepskyblue86:
**What type of PR is this?**
/kind bug

/area plugins

Fix CRI sandbox detection.
When we have an event and we don't have `ContainerStatus`, it's a sandbox.

Fixes #

### DIFF
--- a/plugins/container/go-worker/pkg/container/cri.go
+++ b/plugins/container/go-worker/pkg/container/cri.go
@@ -205,7 +205,6 @@ func (c *criEngine) ctrToInfo(ctx context.Context, ctr *v1.ContainerStatus, podS
 		})
 	}
 
-	isPodSandbox := podSandboxStatus != nil
 	podSandboxID := ctr.Id
 	if podSandboxStatus == nil {
 		podSandboxStatus = &v1.PodSandboxStatus{
@@ -363,7 +362,6 @@ func (c *criEngine) ctrToInfo(ctx context.Context, ctr *v1.ContainerStatus, podS
 			CreatedTime:      nanoSecondsToUnix(ctr.CreatedAt),
 			Env:              ctrInfo.getEnvs(),
 			FullID:           ctr.Id,
-			IsPodSandbox:     isPodSandbox,
 			Labels:           labels,
 			MemoryLimit:      memoryLimit,
 			SwapLimit:        swapLimit,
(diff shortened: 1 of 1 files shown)


plugins/container/go-worker/pkg/container/cri.go
```diff
@@ -520,20 +518,21 @@ func (c *criEngine) Listen(ctx context.Context, wg *sync.WaitGroup) (<-chan even
...
 							ID:          shortContainerID(evt.ContainerId),
 							FullID:      evt.ContainerId,
 							CreatedTime: nanoSecondsToUnix(evt.CreatedAt),
+							IsPodSandbox: true,
 						},
 					}
 				} else {
 					cPodSandbox := evt.GetPodSandboxStatus()
-					podSandboxStatus, _ := c.client.PodSandboxStatus(ctx, cPodSandbox.GetId(), false)
+					podSandboxStatus, _ := c.client.PodSandboxStatus(ctx, cPodSandbox.GetId(), returnInfo)
```
deepskyblue86, line 535: For posterity, `cPodSandbox` and `podSandboxStatus` are available also when it's a sandbox, but we won't use that data for that case.

plugins/container/go-worker/pkg/container/cri.go
```diff
@@ -520,20 +518,21 @@ func (c *criEngine) Listen(ctx context.Context, wg *sync.WaitGroup) (<-chan even
 				}
 
 				var info event.Info
-				// verbose true to return container.Info
-				ctr, err := c.client.ContainerStatus(ctx, evt.ContainerId, true)
+				returnInfo := true
+				ctr, err := c.client.ContainerStatus(ctx, evt.ContainerId, returnInfo)
```
irozzo-1A, line 522: @deepskyblue86 Am I getting it right that we're deducing it's a sandbox just because the call to `ContainerStatus` is failing?

plugins/container/go-worker/pkg/container/cri.go
```diff
@@ -520,20 +518,21 @@ func (c *criEngine) Listen(ctx context.Context, wg *sync.WaitGroup) (<-chan even
...
 				if err != nil || ctr == nil {
 					info = event.Info{
 						Container: event.Container{
 							Type:        c.runtime,
 							ID:          shortContainerID(evt.ContainerId),
 							FullID:      evt.ContainerId,
 							CreatedTime: nanoSecondsToUnix(evt.CreatedAt),
+							IsPodSandbox: true,
```
irozzo-1A, line 530: @deepskyblue86 Am I getting it right that we're deducing it's a sandbox just because the call to ContainerStatus is failing?

deepskyblue86: Exactly. The call fails on a sandbox container (`evt.ContainerId` is the sandbox one, can't have an associated `ContainerStatus`).

PR #1004, opened 2025-10-01
wip: Get container name from annotation for containerd

IanRobertson-wpe:
**What type of PR is this?**

/kind bug

/area plugins

This addresses an issue with the "container" plugin.

The container name is not correctly pulled in when fetching from a containerd source. It uses the shortened container ID instead. This change pulls in the container name from the K8s annotation if it exists and is not empty, otherwise defaulting back to the shortened container ID.

This needs testing.

I identifed this issue in a GKE environment which has a CRI-O socket at /run/containerd/containerd.sock. I had the plugin configured to use CRI-O pointing to /run/crio/crio.sock, and containerd for /run/containerd/containerd.sock. The plugin used the containerd logic for pulling in the data, and this is where I found it was lacking the proper container name, and thus this code contribution. I was able to resolve this issue on my end by disabling containerd and pointing the crio socket to /run/containerd/containerd.sock, but I believe this issue still needs to be addressed for containerd, and I don't have a good way of testing it myself.

### DIFF
--- a/plugins/container/go-worker/pkg/container/containerd.go
+++ b/plugins/container/go-worker/pkg/container/containerd.go
@@ -2,6 +2,10 @@ package container
 
 import (
 	"context"
+	"strconv"
+	"strings"
+	"sync"
+
 	"github.com/containerd/containerd/api/events"
 	containerd "github.com/containerd/containerd/v2/client"
 	"github.com/containerd/containerd/v2/core/containers"
@@ -11,9 +15,6 @@ import (
 	"github.com/falcosecurity/plugins/plugins/container/go-worker/pkg/config"
 	"github.com/falcosecurity/plugins/plugins/container/go-worker/pkg/event"
 	"github.com/opencontainers/runtime-spec/specs-go"
-	"strconv"
-	"strings"
-	"sync"
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


leogr:
Hi @IanRobertson-wpe 

AFAIK, this issue is the same as https://github.com/falcosecurity/falco/issues/3631 and can be solved at the configuration level, see :point_down: 

https://github.com/falcosecurity/charts/pull/892/files#diff-ef3f1a0c0800c919d45ffa98656d27cee2803963e4e13622ac22a08ce6548f0cR418-R424

The rationale is that `containerd` has no concept of a _name_ for containers. Also, it usually exposes two sockets: one using the containerd protocol and another using the CRI protocol. Since CRI supports having a name for containers, the solution is to use CRI to consume `/run/containerd/containerd.sock`.

Please let us know if this works for you as well, or if your case is different.

Thanks!

leogr:
> Hi @IanRobertson-wpe
> 
> AFAIK, this issue is the same as [falcosecurity/falco#3631](https://github.com/falcosecurity/falco/issues/3631) and can be solved at the configuration level, see 👇
> 
> https://github.com/falcosecurity/charts/pull/892/files#diff-ef3f1a0c0800c919d45ffa98656d27cee2803963e4e13622ac22a08ce6548f0cR418-R424
> 
> The rationale is that `containerd` has no concept of a _name_ for containers. Also, it usually exposes two sockets: one using the containerd protocol and another using the CRI protocol. Since CRI supports having a name for containers, the solution is to use CRI to consume `/run/containerd/containerd.sock`.
> 
> Please let us know if this works for you as well, or if your case is different.
> 
> Thanks!

Hey @IanRobertson-wpe 

Have you had a chance to double-check this? If you confirm the issue you encountered is the problem I described above, we can close this PR.

Let us know. Thanks!

IanRobertson-wpe:
@leogr Yes, I can confirm this can be addressed through the configuration adjustments. Closing.

PR #1019, opened 2025-10-09
fix(container): crash at filter extract

deepskyblue86:
**What type of PR is this?**
/kind bug

**Any specific area of the project related to this PR?**
/area plugins

**What this PR does / why we need it**:
I hit a segfault with sinsp-example, just specifying `-f 'container.name!=my_container'`.
After debugging it I realized that event filtering (using extract cap) happens before parse, so the assumption in the code didn't hold.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

PR #1112, opened 2025-12-03
fix: if missing thread_entry, don't attempt to dereference it (iss-1076)

### LINKED ISSUE #1076: bug(container): segfault when `thread_entry` is empty
**Describe the bug**

The container plugin segfaults when cinfo is nil.  Likely this case in `extract.cpp` needs to be changed to `return false`, or the cinfo refs need null checks for each field in the extraction logic.

https://github.com/falcosecurity/plugins/blob/97ba122408c880fc81979d05af81928be37fa089/plugins/container/src/caps/extract/extract.cpp#L603

### Stack of crash
```
Stack trace of thread 1622789:
#0  0x0000000000c7c920 _ZL25dispatch_read_entry_fieldPvS_PKvP20ss_plugin_state_data (falco + 0x87c920)
#1  0x00007f70436b0ade _ZN9my_plugin7extractERKN13falcosecurity20extract_fields_inputE (libcontainer.so + 0x4b0ade)
#2  0x00007f70436b2efd plugin_extract_fields (libcontainer.so + 0x4b2efd)
#3  0x0000000000c5086b _ZN12sinsp_plugin26extract_fields_and_offsetsEP9sinsp_evtjP23ss_plugin_extract_fieldP31ss_plugin_extract_value_offsets (falco + 0x85086b)
#4  0x0000000000c87559 _ZN25sinsp_filter_check_plugin15extract_nocacheEP9sinsp_evtRSt6vectorI15extract_value_tSaIS3_EEPS2_I16extract_offset_tSaIS7_EEb (falco + 0x88>
#5  0x0000000000c378b7 _ZN18sinsp_filter_check20extract_with_offsetsEP9sinsp_evtRSt6vectorI15extract_value_tSaIS3_EEPS2_I16extract_offset_tSaIS7_EEb (falco + 0x8378>
#6  0x0000000000c37950 _ZN18sinsp_filter_check8tostringEP9sinsp_evt (falco + 0x837950)
#7  0x0000000000c1bd8f _ZN19sinsp_evt_formatter19tostring_withformatEP9sinsp_evtRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEENS_13output_formatE (falco + 0x>
#8  0x00000000006f9fb4 _ZNK13falco_formats12format_eventEP9sinsp_evtRKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEES9_S9_S9_RKSt3setIS7_St4lessIS7_ESaIS7_EES>
#9  0x0000000000657351 _ZN13falco_outputs12handle_eventEP9sinsp_evtRKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEES9_N12falco_common13priority_typeES9_RSt3se>
#10 0x0000000000592d7f _ZL24process_inspector_eventsRN5falco3app5stateESt10shared_ptrI5sinspES3_I12stats_writerERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcE>
#11 0x0000000000595143 _ZN5falco3app7actions14process_eventsERNS0_5stateE (falco + 0x195143)
#12 0x000000000052fc42 _ZNSt17_Function_handlerIFN5falco3app10run_resultERNS1_5stateEEPS5_E9_M_invokeERKSt9_Any_dataS4_ (falco + 0x12fc42)
#13 0x000000000052d44d _ZN5falco3app3runERNS0_5stateERbRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE (falco + 0x12d44d)
#14 0x000000000052f7f3 _ZN5falco3app3runEiPPcRbRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE (falco + 0x12f7f3)
#15 0x00000000004b3e58 main (falco + 0xb3e58)
#16 0x00007f70458295d0 __libc_start_call_main (libc.so.6 + 0x295d0)
#17 0x00007f7045829680 __libc_start_main@@GLIBC_2.34 (libc.so.6 + 0x29680)
#18 0x00000000004e39b5 _start (falco + 0xe39b5)

Stack trace of thread 1622791:
#0  0x00007f7043779aa3 n/a (libcontainer.so + 0x579aa3)
#1  0x00007f7043710bf5 n/a (libcontainer.so + 0x510bf5)
#2  0x00007f7043710d29 n/a (libcontainer.so + 0x510d29)
#3  0x00007f7043749749 n/a (libcontainer.so + 0x549749)
#4  0x00007f704373ff95 n/a (libcontainer.so + 0x53ff95)
#5  0x00007f704373fed5 n/a (libcontainer.so + 0x53fed5)
#6  0x00007f7043775ba5 n/a (libcontainer.so + 0x575ba5)
#7  0x00007f704588a19a start_thread (libc.so.6 + 0x8a19a)
#8  0x00007f704590f240 __clone3 (libc.so.6 + 0x10f240)

Stack trace of thread 1622793:
#0  0x00007f7043779aa3 n/a (libcontainer.so + 0x579aa3)
#1  0x00007f7043710ac7 n/a (libcontainer.so + 0x510ac7)
#2  0x00007f704374188c n/a (libcontainer.so + 0x54188c)
#3  0x00007f704374335c n/a (libcontainer.so + 0x54335c)
#4  0x00007f7043744451 n/a (libcontainer.so + 0x544451)
#5  0x00007f70437448c5 n/a (libcontainer.so + 0x5448c5)
#6  0x00007f7043775c30 n/a (libcontainer.so + 0x575c30)
ELF object binary architecture: AMD x86-64
```

**How to reproduce it**

Container event emitted with cinfo nil (missing socket for instance)

**Expected behaviour**

container plugin shouldn't segfault when cinfo is nil or unable to extract metadata

**Screenshots**

**Environment**

- Falco version: 0.42.1

- System info: Linux/x86_64 5.14.0-570.55.1.el9_6.x86_64 Kernel Configuration

- Cloud provider or hardware configuration: N/A
- OS: RHEL9

- Kernel: Linux/x86_64 5.14.0-570.55.1.el9_6.x86_64 Kernel Configuration

- Installation method: RPM, from source

**Additional context**

arirubinstein:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area plugins

> /area registry

> /area build

> /area documentation

Addresses attempted cinfo NPE 

Fixes #1076 

### DIFF
--- a/plugins/container/src/caps/extract/extract.cpp
+++ b/plugins/container/src/caps/extract/extract.cpp
@@ -1,4 +1,5 @@
 #include <plugin.h>
+#include <optional>
 
 //////////////////////////
 // Extract capability
@@ -546,7 +547,7 @@ bool my_plugin::extract(const falcosecurity::extract_fields_input &in)
 
     std::shared_ptr<const container_info> cinfo;
     std::string container_id;
-    falcosecurity::table_entry thread_entry;
+    std::optional<falcosecurity::table_entry> thread_entry;
 
     // Async event. Is that produced by us?
     if(evt_type == PPME_ASYNCEVENT_E)
@@ -575,7 +576,8 @@ bool my_plugin::extract(const falcosecurity::extract_fields_input &in)
             // Retrieve the thread entry associated with this thread id
             thread_entry = m_threads_table.get_entry(tr, thread_id);
             // Retrieve container_id from the entry
-            m_container_id_field.read_value(tr, thread_entry, container_id);
+            m_container_id_field.read_value(tr, thread_entry.value(),
+                                            container_id);
         }
         catch(const std::exception &e)
         {
@@ -763,15 +765,18 @@ bool my_plugin::extract(const falcosecurity::extract_fields_input &in)
     case TYPE_CONTAINER_START_TS:
     case TYPE_CONTAINER_DURATION:
     {
-        uint64_t pidns_init_start_ts;
-        try
+        uint64_t pidns_init_start_ts{0};
+        if(thread_entry.has_value())
         {
-            m_threads_field_pidns_init_start_ts.read_value(tr, thread_entry,
-                                                           pidns_init_start_ts);
-        }
-        catch(...)
-        {
-            pidns_init_start_ts = 0;
+            try
+            {
+                m_threads_field_pidns_init_start_ts.read_value(
+                        tr, thread_entry.value(), pidns_init_start_ts);
+            }
+            catch(...)
+            {
+                pidns_init_start_ts = 0;
+            }
         }
         if(pidns_init_start_ts != 0)
         {
@@ -927,48 +932,61 @@ bool my_plugin::extract(const falcosecurity::extract_fields_input &in)
         break;
     case TYPE_IS_CONTAINER_HEALTHCHECK:
     {
-        int16_t category;
+        int16_t category{CAT_NONE};
         // Since we do write thread category only if not NONE for containerized
         // processes
-        try
+        if(thread_entry.has_value())
         {
-            m_threads_field_category.read_value(tr, thread_entry, category);
-        }
-        catch(...)
-        {
-            category = CAT_NONE;
+            try
+            {
+                m_threads_field_category.read_value(tr, thread_entry.value(),
+                                                    category);
+            }
+            catch(...)
+            {
+                category = CAT_NONE;
+            }
         }
         req.set_value(category == CAT_HEALTHCHECK);
         break;
     }
     case TYPE_IS_CONTAINER_LIVENESS_PROBE:
     {
-        int16_t category;
+        int16_t category{CAT_NONE};
         // Since we do write thread category only if not NONE for containerized
         // processes
-        try
+        if(thread_entry.has_value())
         {
-            m_threads_field_category.read_value(tr, thread_entry, category);
-        }
-        catch(...)
-        {
-            category = CAT_NONE;
+            try
+            {
+                m_threads_field_category.read_value(tr, thread_entry.value(),
+                                                    category);
+            }
+            catch(...)
+            {
+                category = CAT_NONE;
+            }
         }
         req.set_value(category == CAT_LIVENESS_PROBE);
         break;
     }
     case TYPE_IS_CONTAINER_READINESS_PROBE:
     {
-        int16_t category;
+        int16_t category{CAT_NONE};
         // Since we do write thread category only if not NONE for containerized
         // processes
-        try
+        if(thread_entry.has_value())
         {
-            m_threads_field_category.read_value(tr, thread_entry, category);
-        }
-        catch(...)
-        {
-            category = CAT_NONE;
+
+            try
+            {
+                m_threads_field_category.read_value(tr, thread_entry.value(),
+                                                    category);
+            }
+            catch(...)
+            {
+                category = CAT_NONE;
+            }
         }
         req.set_value(category == CAT_READINESS_PROBE);
         break;


deepskyblue86:
> /lgtm
> 
> @deepskyblue86 do you agree?

No, not really. I'm not convinced.
The comment is pretty clear: 
> Can't return anything but those fields without containers metadata.
> Go on to extract other fields if needed, perhaps they'll be one of the above

And those fields won't use `cinfo` (unless we're with a version without https://github.com/falcosecurity/plugins/pull/1019)

arirubinstein:
I'll try to get a reproduction on another machine - is the order of these fields guaranteed from the socket?

arirubinstein:
I think I see the issue now, it's not `cinfo`, but `thread_entry` can potentially be nil, causing the dereference error.  When `is_container_event` is true (lines 567-570), the code sets `cinfo` but never initializes `thread_entry`. Later, when extracting timestamp or probe fields, the uninitialized `thread_entry` is passed to `read_value()`, which internally calls `dispatch_read_entry_field` with invalid memory, causing the segfault.

irozzo-1A:
> No, not really. I'm not convinced. The comment is pretty clear:
> 
> > Can't return anything but those fields without containers metadata.
> > Go on to extract other fields if needed, perhaps they'll be one of the above
> 
> And those fields won't use `cinfo` (unless we're with a version without #1019)

Actually the API of the extract SDK is not very clear to me, but after checking returning `true` or `false` does not really matter if no result was set e.g. `req.set_value()`, thus unless I'm missing something this change cannot be the fix, even if I would prefer returning `false` if no value could be extracted. https://github.com/falcosecurity/libs/blob/0d8e50f47a6135d5f21402bc748ec36fdc1cbd71/userspace/libsinsp/plugin_filtercheck.cpp#L178-L182

irozzo-1A:
> I think I see the issue now, it's not `cinfo`, but `thread_entry` can potentially be nil, causing the dereference error. When `is_container_event` is true (lines 567-570), the code sets `cinfo` but never initializes `thread_entry`. Later, when extracting timestamp or probe fields, the uninitialized `thread_entry` is passed to `read_value()`, which internally calls `dispatch_read_entry_field` with invalid memory, causing the segfault.

@arirubinstein Yes, This analysis seems correct to me and agrees with the stack trace. Thanks!

deepskyblue86:
> Actually the API of the extract SDK is not very clear to me

Same for me, so yesterday I searched for the code: https://github.com/falcosecurity/plugin-sdk-cpp/blob/main/include/falcosecurity/internal/plugin_mixin_extraction.h#L115-L146
The plugin extract is called for each field, when it returns false it will break the loop and exit, so it won't process the other fields.

plugins/container/src/caps/extract/extract.cpp
```diff
@@ -576,6 +577,7 @@ bool my_plugin::extract(const falcosecurity::extract_fields_input &in)
             thread_entry = m_threads_table.get_entry(tr, thread_id);
             // Retrieve container_id from the entry
             m_container_id_field.read_value(tr, thread_entry, container_id);
+            thread_entry_valid = true;
```
irozzo-1A: I would push the `thread_entry` assignment out of this if/else to be able to cover more  cases

plugins/container/src/caps/extract/extract.cpp
```diff
@@ -959,6 +971,11 @@ bool my_plugin::extract(const falcosecurity::extract_fields_input &in)
     }
     case TYPE_IS_CONTAINER_READINESS_PROBE:
     {
+        if(!thread_entry_valid)
+        {
+            req.set_value(false);
```
irozzo-1A: Same here

irozzo-1A:
> Same for me, so yesterday I searched for the code: https://github.com/falcosecurity/plugin-sdk-cpp/blob/main/include/falcosecurity/internal/plugin_mixin_extraction.h#L115-L146 The plugin extract is called for each field, when it returns false it will break the loop and exit, so it won't process the other fields.

Good point @deepskyblue86

plugins/container/src/caps/extract/extract.cpp
irozzo-1A: well, probably it's worthless as the container events should not have a thread ID information associated.

plugins/container/src/caps/extract/extract.cpp
```diff
@@ -575,7 +576,11 @@ bool my_plugin::extract(const falcosecurity::extract_fields_input &in)
             // Retrieve the thread entry associated with this thread id
             thread_entry = m_threads_table.get_entry(tr, thread_id);
             // Retrieve container_id from the entry
-            m_container_id_field.read_value(tr, thread_entry, container_id);
+            if(thread_entry.has_value())
+            {
+                m_container_id_field.read_value(tr, thread_entry.value(),
+                                                container_id);
+            }
```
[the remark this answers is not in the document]
arirubinstein: removed to just `thread_entry.value()`

irozzo-1A:
LGTM @arirubinstein, just fix the formatting issue and it's ready to go 😉

deepskyblue86:
If you didn't setup pre-commit, the patch is as simple as that:
```diff
diff --git a/plugins/container/src/caps/extract/extract.cpp b/plugins/container/src/caps/extract/extract.cpp
index e6d2a00f..0c81c2ac 100644
--- a/plugins/container/src/caps/extract/extract.cpp
+++ b/plugins/container/src/caps/extract/extract.cpp
@@ -576,7 +576,8 @@ bool my_plugin::extract(const falcosecurity::extract_fields_input &in)
             // Retrieve the thread entry associated with this thread id
             thread_entry = m_threads_table.get_entry(tr, thread_id);
             // Retrieve container_id from the entry
-            m_container_id_field.read_value(tr, thread_entry.value(), container_id);
+            m_container_id_field.read_value(tr, thread_entry.value(),
+                                            container_id);
         }
         catch(const std::exception &e)
         {
```

If you did set it up but after the commit ⇒ `pre-commit run --files plugins/container/src/caps/extract/extract.cpp` will do.

arirubinstein:
formatting fixed

deepskyblue86:
I just realized I have no power here [cit.] :sweat_smile: (and that there's no OWNERS file for the container plugin...)

PR #1218, opened 2026-02-19
fix(plugins/container): handle liveness and readiness probes

### LINKED ISSUE #1017: Container plugin: `container.liveness_probe` is always `NONE`
The fields `container.healthcheck`, `container.liveness_probe` and `container.readiness_probe` are always `NONE`. In this case I created a rule that uses these fields in the output. The event type is filtered by `evt.type in (execve, execveat)`.

I captured the process spawning for a few minutes, so it's not a delayed initialization or something like that.

pod:
```
apiVersion: v1
kind: Pod
metadata:
  name: debugging
spec:
  containers:
  - name: netshoot
    image: nicolaka/netshoot
    command: ["/bin/bash", "-c", "while true; do sleep 100; ls; done"]
    livenessProbe:
      exec:
        command:
        - "ls"
        - "-111"
      initialDelaySeconds: 5
      periodSeconds: 10
  - name: alpine
    command: ["/bin/sh", "-c", "while true; do sleep 100; ls; done"]
    image: alpine
  restartPolicy: Always
```

falco.yaml:
```
plugins:
  - init_config: ""
    library_path: libjson.so
    name: json
  - init_config:
      engines:
        bpm:
          enabled: false
        containerd:
          enabled: true
          sockets:
            - /run/containerd/containerd.sock
        cri:
          enabled: true # need to enable it too for some reason (maybe bug?)
          sockets:
            - /run/crio/crio.sock
        docker:
          enabled: false
          sockets:
            - /var/run/docker.sock
        libvirt_lxc:
          enabled: false
        lxc:
          enabled: false
        podman:
          enabled: false
          sockets:
            - /run/podman/podman.sock
      hooks:
        - create
      label_max_len: 100
      with_size: false
    library_path: libcontainer.so
    name: container
```

leogr:
**What type of PR is this?**

/kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area plugins

> /area registry

> /area build

> /area documentation

Fixes #1017

### DIFF
--- a/plugins/container/go-worker/pkg/container/cri.go
+++ b/plugins/container/go-worker/pkg/container/cri.go
@@ -270,6 +270,16 @@ func (c *criEngine) ctrToInfo(ctx context.Context, ctr *v1.ContainerStatus, podS
 		}
 	}
 
+	// Extract liveness/readiness probes from the pod-level annotation
+	var (
+		livenessProbe  *event.Probe
+		readinessProbe *event.Probe
+	)
+	containerName := ctr.GetMetadata().GetName()
+	if annotation, ok := podSandboxStatus.GetAnnotations()[k8sLastAppliedConfigAnnotation]; ok {
+		livenessProbe, readinessProbe = parseProbesFromPodAnnotation(annotation, containerName)
+	}
+
 	var size int64 = -1
 	if config.GetWithSize() {
 		stats, _ := c.client.ContainerStats(ctx, ctr.Id)
@@ -365,6 +375,8 @@ func (c *criEngine) ctrToInfo(ctx context.Context, ctr *v1.ContainerStatus, podS
 			PodSandboxLabels: podSandboxLabels,
 			Mounts:           mounts,
 			Size:             size,
+			LivenessProbe:    livenessProbe,
+			ReadinessProbe:   readinessProbe,
 		},
 	}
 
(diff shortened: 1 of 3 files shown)


leogr:
cc @deepskyblue86

plugins/container/go-worker/pkg/container/cri.go
```diff
@@ -270,6 +270,16 @@ func (c *criEngine) ctrToInfo(ctx context.Context, ctr *v1.ContainerStatus, podS
 		}
 	}
 
+	// Extract liveness/readiness probes from the pod-level annotation
+	var (
+		livenessProbe  *event.Probe
+		readinessProbe *event.Probe
+	)
+	containerName := ctr.GetMetadata().GetName()
+	if annotation, ok := podSandboxStatus.GetAnnotations()[k8sLastAppliedConfigAnnotation]; ok {
```
irozzo-1A: This is a best effort extraction, the [last-applied-configuration](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubectl-kubernetes-io-last-applied-configuration) is part of the legacy mechanism to track changes with client side apply. It's not there when using server-side apply or when creating the resources without apply.

The kubelet exposes an API to list the pods spec managed by it, we could rely on it but it's not always accessible (especially in a secure clutster). Otherwise we could pass from Apiserver proxy.

TBH I would rather deprecate this feature from the container plugin, as it is a Kubernetes specific information and not strictly related to the container runtime.

leogr:
Closing in favor of https://github.com/falcosecurity/plugins/pull/1265

/close

:pray:

PR #1340, opened 2026-05-08
Add a Rust version of the gcpaudit plugin

geraldcombs:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area plugins

/area registry

> /area build

> /area documentation

> /area automation

This is an AI-aided translation of the gcpaudit plugin from Go to Rust. Why do this? Go doesn't officially support loading multiple c-shared libraries, and while this works on most platforms it crashes on macOS on x64, and the Go team has no plans to fix this:

https://github.com/golang/go/issues/65050

Fortunately this doesn't really affect Falco, but it does affect Stratoshark, which loads multiple Falco plugins at startup:

https://gitlab.com/wireshark/wireshark/-/work_items/20869

A separate PR for a Rust version of the cloudtrail plugin can be found at #1350.

I've been shipping this and  plugins with Stratoshark's development installers for macOS and Windows, and so far they seem to work well. I don't have any large capture files available for performance testing, but haven't run into any issues so far. If you have a large cloudtrail or gcpaudit file suitable for testing that you can share I would be grateful.

Questions and outstanding issues:

- cloudtrail_rs and gcpaudit_rs currently have the same IDs and names as their Go counterparts. Should these have different IDs and/or names?

- The cloudtrail Go plugin has a source test, which is absent from the Rust plugin.

https://gitlab.com/wireshark/wireshark/-/work_items/20869

Fixes #


geraldcombs:
The build-packages-aarch64 job is failing with

```
error: rustc 1.88.0 is not supported by the following packages:
  aws-config@1.8.15 requires rustc 1.91.1
  aws-credential-types@1.2.14 requires rustc 1.91.1
  [ ... ]
```

would it be preferable to bump the rustc version to 1.91.1 or downgrade the dependency versions?

ekoops:
Hey @geraldcombs . Thank you for this contribution. I like the idea, and it for sure will let us avoid the problem you mentioned and any other oddity linked to the go runtime.
I don't know what other mantainers feel about it, but to me it's perfectly fine to replace the original plugins with these, once they'll reach a reasonable shape (so keeping the same IDs is fine).
However, if we reach the consensus on moving forward with this replacement (or even merging this), I would at least split this PR into two: one for `cloudtrail` and the other one for `gcpaudit`. It is really difficult to spot any issue in a 10k PR :joy:

geraldcombs:
> I would at least split this PR into two: one for `cloudtrail` and the other one for `gcpaudit`. It is really difficult to spot any issue in a 10k PR 😂

Great suggestion! Done in #1350.

PR #1355, opened 2026-05-17
fix(plugins/gcpaudit): populate gcp.policyDelta, which was empty for all resource types

### LINKED ISSUE #1351: bug(gcpaudit) gcp.policyDelta: add additional resource types
**Describe the bug**

gcp.policyDelta does not use the correct JSON path with resource types `project`, `folder`, `organization`, `service_account`.   The `else` path is really only used by `bigquery_dataset` type.

The quick fix is to change `plugins/gcpaudit/pkg/gcpaudit/extract.go` line 96 to:

`		if resource == "gcs_bucket" || resource == "project" || resource == "folder" || resource == "organization" || resource == "service_account" {`

**How to reproduce it**

Create a rule to match `gcp.methodName="setIamPoilcy" and jevt.value[/resource/type]="service_account"`.  Add %gcp.policyDelta to the output.  Impersonate a service account in GCP.  The event log output for gcp.policyDelta will be blank because it is trying to match `/protoPayload/metadata/datasetChange/bindingDeltas`, which does not exist in the log.

**Expected behaviour**

gcp.policyDelta matches `/protoPayload/serviceData/policyDelta/bindingDeltas` when /resource/type in (`project`, `folder`, `organization`, `service_account`, `gcs_bucket`).

raajheshkannaa:
## What type of PR is this?

/kind bug

## Any specific area of the PR?

/area plugins

## What this PR does / why we need it

`gcp.policyDelta` returns empty on every resource type. There are two causes.

**1. Serialization.** The field resolves to the `bindingDeltas` array, but the generic tail of `Extract` calls `req.SetValue(string(fsval.GetStringBytes()))`. `GetStringBytes` returns nil for any value that isn't a JSON string, so an array always serialized to `""`. This affected `gcs_bucket` too, the one resource type the old allowlist routed to the correct path:

```
UPSTREAM: gcs_bucket gcp.policyDelta = ""
FIXED:    gcs_bucket gcp.policyDelta = "[{\"action\":\"ADD\",\"role\":\"roles/storage.objectViewer\",\"member\":\"allUsers\"}]"
```

Fixed by serializing with `MarshalTo` and setting the value directly, the same treatment `gcp.resourceLabels` already gets.

**2. Path selection.** The old code sent everything except `gcs_bucket` to BigQuery's `metadata.datasetChange`, so `project`, `folder`, `organization` and `service_account` looked in a location that doesn't exist for them. That is the symptom reported in #1351.

Rather than extend the allowlist, the deltas are now located by payload shape:

```go
var policyDeltaPaths = [][]string{
	{"protoPayload", "serviceData", "policyDelta", "bindingDeltas"},
	{"protoPayload", "metadata", "datasetChange", "bindingDeltas"},
	{"protoPayload", "metadata", "policyDelta", "bindingDeltas"},
}
```

First non-empty array wins. New resource types work without a code change, and a `serviceData` that arrives stripped or empty falls through to `metadata` instead of yielding an empty field. The third path covers services that migrate the same `google.iam.v1.logging.AuditData` proto to the newer `metadata` field, per @leogr's review.

## Testing

`extract_test.go` drives `Extract` end to end through a fake `sdk.ExtractRequest` and `sdk.EventReader`, following the `testExtractRequest` pattern in `plugins/k8saudit`. Covering `Extract` rather than the path lookup alone is what surfaced the serialization bug.

Fixtures: `gcs_bucket` and project-level `serviceData.policyDelta`, `bigquery_dataset` `metadata.datasetChange`, and `service_account` IAM `AuditData` under `metadata.policyDelta`. Negative cases assert an empty field and no error for a stripped `serviceData` wrapper and an unrelated event. `TestPolicyDeltaPathPriority` pins the probe order.

All three positive cases fail against `main` and pass with this change.

The fixtures are **synthetic**, modelled on the `SetIamPolicy` sample in [Google's audit log reference](https://docs.cloud.google.com/logging/docs/audit/understanding-audit-logs#sample) and the shapes described in #1351. They are not captured production logs, and I don't have a GCP org with an exporting Pub/Sub sink to capture one. They prove routing and serialization, not what a real exported event carries on the wire.

## Which issue(s) this PR fixes

Fixes #1351

```release-note
fix(plugins/gcpaudit): gcp.policyDelta returned an empty value for every resource type. The bindingDeltas array is now serialized to JSON, and its location is resolved by payload shape so project, folder, organization and service_account events are populated.
```

### DIFF
--- a/plugins/gcpaudit/pkg/gcpaudit/extract.go
+++ b/plugins/gcpaudit/pkg/gcpaudit/extract.go
@@ -92,12 +92,7 @@ func (p *Plugin) Extract(req sdk.ExtractRequest, evt sdk.EventReader) error {
 
 	case "gcp.policyDelta":
 		resource := string(p.jdata.Get("resource").Get("type").GetStringBytes())
-
-		if resource == "gcs_bucket" {
-			fsval = p.jdata.Get("protoPayload", "serviceData", "policyDelta", "bindingDeltas")
-		} else {
-			fsval = p.jdata.Get("protoPayload", "metadata", "datasetChange", "bindingDeltas")
-		}
+		fsval = p.jdata.Get(policyDeltaPath(resource)...)
 
 	case "gcp.methodName":
 		fsval = p.jdata.Get("protoPayload", "methodName")
@@ -200,4 +195,18 @@ func (p *Plugin) Extract(req sdk.ExtractRequest, evt sdk.EventReader) error {
 	}
 
 	return nil
+}
+
+// policyDeltaPath returns the JSON path that contains the bindingDeltas array
+// for a SetIamPolicy audit event on the given resource type. The generic IAM
+// AuditData proto (google.iam.v1.PolicyDelta) is emitted under
+// protoPayload.serviceData.policyDelta for every resource type that uses the
+// shared IAM service. BigQuery's dataset.setIamPolicy is the outlier and emits
+// its deltas under protoPayload.metadata.datasetChange as part of the
+// BigQueryAuditMetadata. See issue #1351.
+func policyDeltaPath(resource string) []string {
+	if resource == "bigquery_dataset" {
+		return []string{"protoPayload", "metadata", "datasetChange", "bindingDeltas"}
+	}
+	return []string{"protoPayload", "serviceData", "policyDelta", "bindingDeltas"}
 }
--- a/plugins/gcpaudit/pkg/gcpaudit/extract.go
+++ b/plugins/gcpaudit/pkg/gcpaudit/extract.go
@@ -91,8 +91,17 @@ func (p *Plugin) Extract(req sdk.ExtractRequest, evt sdk.EventReader) error {
 		fsval = p.jdata.Get("protoPayload", "request")
 
 	case "gcp.policyDelta":
-		resource := string(p.jdata.Get("resource").Get("type").GetStringBytes())
-		fsval = p.jdata.Get(policyDeltaPath(resource)...)
+		fsval = policyDelta(p.jdata)
+		if fsval != nil {
+			bindingDeltas := fsval.MarshalTo(nil)
+			if len(bindingDeltas) > 0 {
+				req.SetValue(string(bindingDeltas))
+				if req.WantOffset() {
+					req.SetValueOffset(sdk.PluginEventPayloadOffset+uint32(fsval.Offset()), uint32(fsval.Len()))
+				}
+			}
+		}
+		return nil
 
 	case "gcp.methodName":
 		fsval = p.jdata.Get("protoPayload", "methodName")
@@ -170,7 +179,7 @@ func (p *Plugin) Extract(req sdk.ExtractRequest, evt sdk.EventReader) error {
 			if len(resourceLabels) > 0 {
 				req.SetValue(string(resourceLabels))
 				if req.WantOffset() {
-					req.SetValueOffset(sdk.PluginEventPayloadOffset + uint32(fsval.Offset()), uint32(fsval.Len()))
+					req.SetValueOffset(sdk.PluginEventPayloadOffset+uint32(fsval.Offset()), uint32(fsval.Len()))
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


plugins/gcpaudit/pkg/gcpaudit/extract.go
```diff
@@ -201,3 +196,17 @@ func (p *Plugin) Extract(req sdk.ExtractRequest, evt sdk.EventReader) error {
...
 	return nil
 }
+
+// policyDeltaPath returns the JSON path that contains the bindingDeltas array
+// for a SetIamPolicy audit event on the given resource type. The generic IAM
+// AuditData proto (google.iam.v1.PolicyDelta) is emitted under
+// protoPayload.serviceData.policyDelta for every resource type that uses the
+// shared IAM service. BigQuery's dataset.setIamPolicy is the outlier and emits
+// its deltas under protoPayload.metadata.datasetChange as part of the
+// BigQueryAuditMetadata. See issue #1351.
```
leogr: Nice improvement over the old logic 🙏 defaulting to `serviceData.policyDelta` and special-casing BigQuery is the right direction, and I don't see a breaking change here (`gcs_bucket` and `bigquery_dataset` keep the same paths).

A couple of caveats about this rationale though:

- `protoPayload.serviceData` is officially deprecated by Google (they recommend `metadata` now), and it's shared by several services, not just BigQuery. So "BigQuery is the outlier" is a bit of an oversimplification - not a blocker, but the comment may not hold as more services migrate to `metadata`.

- More importantly, this plugin ingests *exported* logs via the Pub/Sub sink, and `serviceData` is known to be stripped on exported entries even when it's populated in Logs Explorer 👉 https://permiso.io/blog/gcp-servicedata-officially-deprecated-actively-dangerous

So for the very resources #1351 is about (project/folder/org/service_account), `gcp.policyDelta` may still come back empty in practice after this change. Have you confirmed it's populated on a real exported `setIamPolicy` event? 🤔 If not, we may need `metadata` as a fallback.

plugins/gcpaudit/pkg/gcpaudit/extract_test.go
```diff
@@ -0,0 +1,57 @@
...
+		// Unknown / future resource types default to the generic IAM path,
+		// not the BigQuery-specific datasetChange path.
+		{"unknown_type", serviceData},
+		{"", serviceData},
+		{"bigquery_dataset", datasetChange},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.resource, func(t *testing.T) {
+			got := policyDeltaPath(tc.resource)
```
leogr: The test covers the path selection nicely, but it stops at `policyDeltaPath()` and never exercises `Extract()` with a real event, so it won't catch the case where the selected path exists but is empty on the wire (see my comment on `extract.go`).

May you add a fixture-based test that feeds a captured exported `setIamPolicy` event (e.g. project-level) through `Extract` and asserts `gcp.policyDelta` is non-empty? That would guard the actual #1351 symptom, not just the routing. 🙏

plugins/gcpaudit/pkg/gcpaudit/extract.go
raajheshkannaa: Thanks, this was the right thing to push on. Chasing it down turned up a second, bigger problem, and you were closer to it than I was.

**`gcp.policyDelta` was empty for every resource type, including `gcs_bucket`.**

The field resolves to the `bindingDeltas` array, but the generic tail of `Extract` ends in `req.SetValue(string(fsval.GetStringBytes()))`. `GetStringBytes` returns nil for anything that isn't a JSON string, so an array always serialized to `""`. Path selection was necessary but never sufficient. I confirmed it by running a `gcs_bucket` fixture, the one type the old allowlist routed correctly, against `main`:

```
UPSTREAM: gcs_bucket gcp.policyDelta = ""
FIXED:    gcs_bucket gcp.policyDelta = "[{\"action\":\"ADD\",\"role\":\"roles/storage.objectViewer\",\"member\":\"allUsers\"}]"
```

Fixed by serializing with `MarshalTo` and setting the value directly, the same treatment `gcp.resourceLabels` already gets.

**On `serviceData` and the export gap.**

Your caveat about the deprecation was fair, and the comment I wrote was too confident. Rewritten. But I don't think the Permiso findings undercut the `serviceData` path here, for two reasons:

1. That post's stripped example is `auditConfigDeltas`, the audit-config sibling. Its own counter-example is a role binding `SetIamPolicy` from the same `cloudresourcemanager.googleapis.com` that exported with `serviceData` intact. `policyDelta` binding changes are the case that survived in their testing.
2. Google's audit log reference still documents `serviceData.policyDelta.bindingDeltas` in its canonical `SetIamPolicy` sample, and still lists IAM and Cloud Storage under [service-specific audit data](https://docs.cloud.google.com/logging/docs/audit/understanding-audit-logs#servicedata-services) as using `google.iam.v1.logging.AuditData`. As that post notes, the list has been growing, not shrinking.

The reporter on #1351 also observed the deltas at the `serviceData` path on their own logs.

**To answer your direct question: no, I have not confirmed this against a real exported event.** I don't have a GCP org with an exporting Pub/Sub sink to capture one, so I'm not going to claim otherwise. Everything above is from the documented proto shapes, the issue report, and the post you linked.

Given that, I took your `metadata` fallback suggestion rather than betting on one path. The lookup now probes by payload shape instead of switching on `resource.type`:

```go
var policyDeltaPaths = [][]string{
	{"protoPayload", "serviceData", "policyDelta", "bindingDeltas"},
	{"protoPayload", "metadata", "datasetChange", "bindingDeltas"},
	{"protoPayload", "metadata", "policyDelta", "bindingDeltas"},
}
```

First non-empty array wins. A stripped or empty `serviceData` falls through to `metadata` instead of yielding an empty field, new resource types need no code change, and the third entry covers services that migrate the same IAM `AuditData` proto to `metadata`. I've flagged that last path in the comment as forward-compat; I haven't seen it in the wild.

If you'd rather I hold this until someone can attach a captured exported event, I'm fine with that. Happy to keep it open.

plugins/gcpaudit/pkg/gcpaudit/extract_test.go
raajheshkannaa: Done, and this is the comment that found the real bug. Testing `policyDeltaPath` in isolation was exactly the gap you described: the path was right and the field was still empty, because the array never survived `GetStringBytes`.

`extract_test.go` now drives `Extract` end to end through a fake `sdk.ExtractRequest` and `sdk.EventReader`, following the `testExtractRequest` pattern already in `plugins/k8saudit/pkg/k8saudit/extract_test.go`. Fixtures cover four shapes:

- `gcs_bucket` with `serviceData.policyDelta` (regression guard, the path the old code got right and still returned empty on)
- project-level `SetIamPolicy` with `serviceData.policyDelta`
- `bigquery_dataset` with `metadata.datasetChange`
- `service_account` with the IAM `AuditData` under `metadata.policyDelta`

Plus negative cases: a stripped `serviceData` wrapper with no `policyDelta`, and an unrelated event. Both assert an empty field and no error rather than a partial value. `TestPolicyDeltaPathPriority` pins the probe order so an empty `serviceData` array can't shadow populated `datasetChange` deltas.

All three positive cases fail against `main`:

```
--- FAIL: TestExtractPolicyDelta/gcs_bucket_serviceData
--- FAIL: TestExtractPolicyDelta/bigquery_datasetChange
--- FAIL: TestExtractPolicyDelta/IAM_AuditData_under_metadata
        gcp.policyDelta is empty; this is the #1351 symptom
```

One caveat I want to be explicit about: **these fixtures are synthetic**, not captured production logs. They're modelled on the `SetIamPolicy` sample in Google's audit log reference and the shapes described in #1351, trimmed to the fields the plugin reads. The file comment says so. They prove the routing and the serialization; they can't prove what a real exported event contains on the wire. If someone can attach a captured one, swapping it in is a small change.

Two things I found while in here and deliberately left alone, since they're outside #1351:

1. `gcp.authorizationInfo` (array) and `gcp.request` (object) have the identical serialization defect. Both are described as `(JSON)` in `Fields()` and both go through `GetStringBytes`, so both should be returning empty. I haven't verified them the way I did `policyDelta`. Say the word and I'll fix them here or open a separate issue.
2. `Extract` skips parsing when `evt.EventNum() == p.lastEventNum`, and `lastEventNum`'s zero value is `0`. On a fresh plugin, an event numbered 0 would never be parsed. I don't know whether the framework ever emits event 0, so this may be unreachable. The test helper starts at 1 to avoid depending on it.

leogr (approved):
Hey @raajheshkannaa, thanks for the rework, and for chasing this down to the real bug 🙏

I went through the new head and it looks good to me:

- Both points from my first round are addressed: the comment no longer paints BigQuery as the outlier, and `metadata` is now probed as a fallback when `serviceData` is missing or stripped on export.
- The serialization fix is correct: `GetStringBytes` returns nil for anything that is not a JSON string, so the array always came back as `""`. Good catch.
- Our CI does not run `go test` for the Go plugins yet, so I ran them locally on your branch: they pass.

A few notes, none of them blocking:

- Tide rebases in this repo (no squash), so both commits land on `main` as they are, and the first commit message still describes the `resource.type` routing that the second commit replaced. If you feel like squashing into one commit, that would keep the history and the generated changelog clean. Up to you.
- `gcp.authorizationInfo` and `gcp.request`: same root cause, as you noted. Let's keep this PR scoped as is 👉 please open a separate issue (and a PR if you want to take it), happy to review it.
- Synthetic fixtures are fine for this fix. If someone attaches a real exported `SetIamPolicy` event later, swapping it in is cheap.

LGTM

PR #1400, opened 2026-06-16
fix(k8saudit): extract per-container fields from init and ephemeral containers

kanywst:
**What type of PR is this?**

/kind bug

/area plugins

Add `ka.req.pod.initContainers.*` and `ka.req.pod.ephemeralContainers.*`, keeping `ka.req.pod.containers.*` on `spec.containers`. Create Privileged Pod now matches any container kind through an `any_container_privileged` macro, so a privileged init or ephemeral container is no longer missed by the shipped rules.

```release-note
new(k8saudit): add ka.req.pod.initContainers.* and ka.req.pod.ephemeralContainers.* fields and cover privileged init/ephemeral containers in the rules
```


plugins/k8saudit/pkg/k8saudit/container_lists_test.go
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: Apache-2.0
+/*
+Copyright (C) 2023 The Falco Authors.
```
leogr:
```suggestion
Copyright (C) 2026 The Falco Authors.
```

leogr (commented):
Hey @kanywst

First of all, thanks for digging into this, and welcome 🙏

The gap is real since init and ephemeral containers carry their own securityContext, however, I'm not convinced about reading them through ka.req.pod.containers.*. 

In the K8s pod spec, `containers`, `initContainers` and `ephemeralContainers` are three distinct arrays, so merging them makes the field mean more than its name says. 

It also changes the index: `ka.req.pod.containers.image[0]` used to be "the first container's image", but with the merge, `[0]` returns the first of each list, so it no longer points at a single container.


I believe a cleaner approach would be:
  1. Leave `ka.req.pod.containers.*` as is (only spec.containers).
  2. Add `ka.req.pod.initContainers.*` and `ka.req.pod.ephemeralContainers.*`. (The least invasive way is to add a list parameter to the existing readContainer* helpers instead of duplicating the switch cases.  N.B. ephemeral containers can't have ports, so we can skip `host_port` for that family.)
  3. Keep the "any container kind" union in the rules, via a macro:
  ```
  - macro: any_container_privileged
     condition: >
       (ka.req.pod.containers.privileged intersects (true) or
        ka.req.pod.initContainers.privileged intersects (true) or
        ka.req.pod.ephemeralContainers.privileged intersects (true))
  ```

Then "Create Privileged Pod" (and the couple of rules with an image allowlist) would use the macro. As a bonus, the "EphemeralContainers Created" rule could drop its %jevt.value[/requestObject/ephemeralContainers/0/name] JSON pointer for the new typed fields.

This keeps the fields faithful to the API and the "treat all containers alike" choice in the rules, where it belongs. It is a bit more work than the current patch (a few more fields, a small rules pass, and `make readme` to regenerate the table), but it is definitely feasible.

Thanks again, and happy to help along the way 👍

ekoops:
Totally agree with @leogr . Maintaining the separation enables more flexibility and accuracy

kanywst:
@leogr 
Thanks for the detailed review. Reworked along those lines: containers.* stays on spec.containers; initContainers and ephemeralContainers get their own fields (no host_port on ephemeral), share one handler, and Create Privileged Pod uses an any_container_privileged macro.

plugins/k8saudit/rules/k8s_audit_rules.yaml
```diff
@@ -237,10 +237,17 @@
...
+    (ka.req.pod.containers.privileged intersects (true) or
+     ka.req.pod.initContainers.privileged intersects (true) or
+     ka.req.pod.ephemeralContainers.privileged intersects (true))
+
 - rule: Create Privileged Pod
   desc: >
     Detect an attempt to start a pod with a privileged container
-  condition: kevt and pod and kcreate and ka.req.pod.containers.privileged intersects (true) and not ka.req.pod.containers.image.repository in (k8s_audit_privileged_images)
+  condition: kevt and pod and kcreate and any_container_privileged and not ka.req.pod.containers.image.repository in (k8s_audit_privileged_images)
   output: Pod started with privileged container (user=%ka.user.name pod=%ka.resp.name resource=%ka.target.resource ns=%ka.target.namespace images=%ka.req.pod.containers.image)
```
leogr:
```suggestion
  output: Pod started with privileged container (user=%ka.user.name pod=%ka.resp.name resource=%ka.target.resource ns=%ka.target.namespace images=%ka.req.pod.containers.image init_images=%ka.req.pod.initContainers.image ephemeral_images=%ka.req.pod.ephemeralContainers.image)
```

The condition now matches a privileged container in any of the three lists, but the output still prints only `containers` images. So when an init or ephemeral container is the privileged one, the alert shows the wrong image (or none). This adds the init/ephemeral images so the alert shows what actually triggered it.

leogr (commented):
Hey @kanywst 

Thanks! Everything SGMT to me now, just one nit. See my comment below 🙏

plugins/k8saudit/rules/k8s_audit_rules.yaml
kanywst: Thanks. Applied.

leogr (approved):

PR #1467, opened 2026-08-03
update(deps): migrate to go.podman.io/podman/v6, bump containerd to v2.2.5

leogr:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

/kind failing-test

> /kind feature

/area plugins

> /area registry

> /area build

> /area documentation

> /area automation

containerd v2.2.5 requires opencontainers/runtime-spec v1.3.0, which changed LinuxPids.Limit from int64 to *int64. podman/v5 still assigns an int64 there, up to and including the latest v5.8.5, so the two cannot be built together:

    podman/v5@v5.8.2/pkg/specgen/resources_linux.go:17:13: cannot use limit (variable of type int64) as *int64 value in struct literal

Holding runtime-spec back at v1.2.1 is not an option, since containerd v2.2.5 and runtime-tools already expect the pointer form.

podman v6 is built against runtime-spec v1.3.0 and moves to the go.podman.io namespace following the CNCF org transfer, as already done for containers/image in a167a1ca. The supported server floor is unchanged: v5.8.2 and v6.0.2 both declare a minimal Libpod API version of 4.0.0, so the documented podman >= v4.0.0 requirement still holds.

podman v6 surfaces moby/moby event types instead of docker/docker ones, so podman.go switches its events import accordingly. docker.go keeps docker/docker, which is still correct for the Docker engine.

Fixes #


ekoops (approved):
/approve

PR #1501, opened 2026-08-27
fix(plugins/container): link libresolv on Linux so the plugin loads on glibc < 2.34

### LINKED ISSUE #1500: container: libcontainer.so fails to load on glibc < 2.34 (undefined symbol: __res_search) — missing -lresolv on Linux
## Describe the bug

`libcontainer.so` (the container plugin bundled with every Falco release since ~0.42.0) fails to `dlopen` on any host with glibc older than 2.34:

```
Runtime error: cannot load plugin /usr/share/falco/plugins/libcontainer.so:
can't load plugin dynamic library: /usr/share/falco/plugins/libcontainer.so:
undefined symbol: __res_search. Exiting.
```

Because the default `falco_rules.yaml` declares `required_plugin_versions: container`, the failure is total: **stock Falco cannot start at all with its shipped default configuration** on these distros. Under systemd it crash-loops indefinitely.

Affected platforms include currently supported enterprise distros: Debian 11 (glibc 2.31), RHEL/Rocky/AlmaLinux 8 (2.28), Ubuntu 20.04 (2.31), Ubuntu 18.04 (2.27), Amazon Linux 2 (2.26).

Previously reported against the falco repo as falcosecurity/falco#3719 (AlmaLinux 8.10, Falco 0.42.0) and falcosecurity/falco#3728 — both went stale and were closed unfixed. Verified still present in **Falco 0.44.1 / container plugin 0.7.1**, reproduced on Debian 11 and Rocky Linux 8.

## Root cause

The go-worker static library uses the Go `net` package, whose cgo resolver calls `res_search` from libresolv (Go >= 1.20). `cmake/modules/go-worker.cmake` handles this **only for macOS**:

```cmake
if(APPLE)
    find_library(SECURITY_FRAMEWORK Security REQUIRED)
    find_library(RESOLV resolv REQUIRED)
    find_library(CORE CoreFoundation REQUIRED)
    set(WORKER_DEP ${SECURITY_FRAMEWORK} ${RESOLV} ${CORE})
endif()
```

There is no Linux equivalent, so the shipped `.so` carries an undefined `__res_search` with **no `DT_NEEDED` entry for `libresolv.so.2`**:

```
$ readelf -d /usr/share/falco/plugins/libcontainer.so | grep NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]

$ readelf --dyn-syms -W /usr/share/falco/plugins/libcontainer.so | grep -w UND | grep res_search
   113: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND __res_search
```

On glibc >= 2.34, libresolv is merged into `libc.so.6`, which exports `__res_search` as a compat symbol — masking the missing dependency on all modern build/test hosts (the release CI builds on `debian:bullseye` precisely to keep old-glibc compatibility, but this defeats it). On older glibc the symbol lives only in `libresolv.so.2`, which is never loaded, so `dlopen` fails.

**Confirmation experiment** (Debian 11, stock Falco 0.44.1): the identical command run twice —

```
$ falco -o engine.kind=modern_ebpf
Runtime error: cannot load plugin ... undefined symbol: __res_search. Exiting.

$ LD_PRELOAD=/lib/x86_64-linux-gnu/libresolv.so.2 falco -o engine.kind=modern_ebpf
... Opening 'syscall' source with modern BPF probe.   # starts cleanly
```

Forcing exactly the library a `DT_NEEDED` entry would have loaded flips broken to working. (The `LD_PRELOAD` also serves as a stopgap for affected deployments, via a systemd drop-in.)

## How to reproduce it

On any Debian 11 / RHEL 8-family / Ubuntu 20.04 host:

1. Install Falco 0.44.1 from the official apt/yum repo
2. Start any engine flavor (`falco-modern-bpf.service`, `falco-bpf.service`, or `falco-kmod.service`), or run `falco` directly
3. Observe the plugin load failure and exit; under systemd, a crash loop

## Expected behaviour

Falco starts with its default configuration; the container plugin loads on all supported glibc versions.

## Fix (tested)

Two small CMake changes, verified end to end:

1. `cmake/modules/go-worker.cmake` — add a Linux branch mirroring the APPLE one: `set(WORKER_DEP resolv)`. `libresolv.so.2` exists on every glibc (real library pre-2.34, compat stub after), so the added `DT_NEEDED` is safe everywhere.
2. `CMakeLists.txt` — link `${WORKER_LIB} ${WORKER_DEP}` (archive before library) instead of the current `${WORKER_DEP} ${WORKER_LIB}`. With the current order GNU ld processes `-lresolv` before anything references its symbols and discards it, so fix (1) alone silently doesn't take. (ld64 is order-insensitive, which is why the macOS branch works as-is.)

Verification performed with plugin v0.7.1 rebuilt in `debian:bullseye` (the release CI's own baseline):

- rebuilt `.so` gains `NEEDED: libresolv.so.2`; the symbol becomes properly versioned (`__res_search@GLIBC_2.2.5` on x86_64, `@GLIBC_2.17` on aarch64)
- **Debian 11 (glibc 2.31)**: stock Falco 0.44.1 with only the plugin binary replaced — default ruleset validates, service runs under systemd, rules fire, and events are enriched (`container_id=host ...`)
- **Ubuntu 24.04 (glibc 2.39)**: patched plugin behaves identically to stock — no regression
- **Linux aarch64**: builds clean with the same fix in the dynamic table
- macOS/Windows: untouched (APPLE branch unchanged; `WORKER_DEP` unset on Windows)

PR with this change follows.

## Environment

- Falco version: 0.44.1 (also reproduced on 0.42.x per the earlier falco-repo issues)
- Container plugin version: 0.7.1
- System info: Debian 11 (kernel 5.10.0-46-amd64, glibc 2.31) and Rocky Linux 8.10 (kernel 4.18.0-553, glibc 2.28), amd64
- Installation method: official apt/yum packages

ManuelFCastillo:
**What type of PR is this?**

/kind bug

/area plugins

Makes `libcontainer.so` loadable on hosts with glibc < 2.34 (Debian 11, RHEL/Rocky/AlmaLinux 8, Ubuntu 20.04, ...), where it currently fails with `undefined symbol: __res_search` — which, combined with the default Falco ruleset requiring the container plugin, prevents stock Falco from starting at all on those distros.

The go-worker library's cgo resolver references `res_search` from libresolv (Go >= 1.20), but the resolv link dependency was only declared in the `if(APPLE)` branch of `go-worker.cmake`. The shipped Linux `.so` therefore has an undefined `__res_search` and no `DT_NEEDED` entry for `libresolv.so.2`. On glibc >= 2.34 the symbol resolves via libc's compat exports, masking the bug on modern hosts; on older glibc it lives only in `libresolv.so.2`, which is never loaded.

Two changes (both linkage-only, no code changes):

1. `cmake/modules/go-worker.cmake`: set `WORKER_DEP` to `resolv` on Linux, mirroring the existing APPLE branch. `libresolv.so.2` exists on all glibc versions (real library before 2.34, compat stub after), so the added dependency is safe everywhere.
2. `CMakeLists.txt`: link `${WORKER_LIB} ${WORKER_DEP}` (archive before its dependencies). With the previous order, GNU ld processes `-lresolv` before anything references its symbols and discards it, so change (1) alone silently has no effect. ld64 is order-insensitive, which is why the macOS branch works as-is.

Also bumps the plugin version to 0.7.2.

Verification performed (details and full logs in the linked issue):

- Rebuilt v0.7.1 with this patch in `debian:bullseye` (the release CI baseline): the `.so` gains `NEEDED: libresolv.so.2` and the symbol becomes properly versioned (`__res_search@GLIBC_2.2.5` on x86_64, `@GLIBC_2.17` on aarch64).
- Debian 11 (glibc 2.31): stock Falco 0.44.1 with only the plugin binary replaced — default ruleset validates, service runs under systemd, rules fire, events are enriched. Before the patch: immediate crash loop.
- Ubuntu 24.04 (glibc 2.39): patched plugin behaves identically to stock — no regression.
- Linux aarch64: builds clean with the same result.
- The `LD_PRELOAD=libresolv.so.2` experiment in the issue independently confirms the missing-DT_NEEDED mechanism on the shipped binary.

Fixes #1500

(Also addresses the symptoms previously reported in falcosecurity/falco#3719 and falcosecurity/falco#3728, which went stale.)

The link-order half of the fix is easy to lose: if `-lresolv` is added but placed before the go-worker archive, GNU ld silently drops it and the resulting `.so` is byte-for-byte as broken as before. `readelf -d libcontainer.so | grep resolv` is the quick check that the fix actually took.

```release-note
fix(plugins/container): link libresolv on Linux so libcontainer.so loads on hosts with glibc < 2.34 (Debian 11, RHEL/Rocky/Alma 8, Ubuntu 20.04)
```

### DIFF
--- a/plugins/container/CMakeLists.txt
+++ b/plugins/container/CMakeLists.txt
@@ -9,7 +9,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
 # project metadata
 project(
         container
-        VERSION 0.7.1
+        VERSION 0.7.2
         DESCRIPTION "Falco container metadata enrichment Plugin"
         LANGUAGES CXX)
 
@@ -65,7 +65,7 @@ endif()
 target_include_directories(container PRIVATE ${CMAKE_BINARY_DIR}/src/ src/ ${PLUGIN_SDK_INCLUDE} ${PLUGIN_SDK_DEPS_INCLUDE} ${WORKER_INCLUDE})
 
 # project linked libraries
-target_link_libraries(container PRIVATE fmt::fmt-header-only ReflexLibStatic ${WORKER_DEP} ${WORKER_LIB})
+target_link_libraries(container PRIVATE fmt::fmt-header-only ReflexLibStatic ${WORKER_LIB} ${WORKER_DEP})
 
 option(ENABLE_TESTS "Enable build of unit tests" ON)
 
--- a/plugins/container/cmake/modules/go-worker.cmake
+++ b/plugins/container/cmake/modules/go-worker.cmake
@@ -25,6 +25,15 @@ if(APPLE)
     find_library(RESOLV resolv REQUIRED)
     find_library(CORE CoreFoundation REQUIRED)
     set(WORKER_DEP ${SECURITY_FRAMEWORK} ${RESOLV} ${CORE})
+elseif(CMAKE_HOST_SYSTEM_NAME STREQUAL "Linux")
+    # The same Go >= 1.20 cgo-resolver requirement applies on Linux: the net
+    # package references res_search from libresolv. On glibc >= 2.34 the
+    # symbol happens to resolve through libc's compat exports, which masks a
+    # missing dependency; on older glibc (Debian 11, RHEL/Rocky/Alma 8,
+    # Ubuntu 20.04) it lives only in libresolv.so.2, and without a DT_NEEDED
+    # entry the plugin fails to dlopen with:
+    #   undefined symbol: __res_search
+    set(WORKER_DEP resolv)
 endif()
 set(WORKER_LIB ${CMAKE_SOURCE_DIR}/go-worker/libworker.a)
 set(WORKER_INCLUDE ${CMAKE_SOURCE_DIR}/go-worker)


plugins/container/CMakeLists.txt
```diff
@@ -65,7 +65,7 @@ endif()
 target_include_directories(container PRIVATE ${CMAKE_BINARY_DIR}/src/ src/ ${PLUGIN_SDK_INCLUDE} ${PLUGIN_SDK_DEPS_INCLUDE} ${WORKER_INCLUDE})
 
 # project linked libraries
-target_link_libraries(container PRIVATE fmt::fmt-header-only ReflexLibStatic ${WORKER_DEP} ${WORKER_LIB})
+target_link_libraries(container PRIVATE fmt::fmt-header-only ReflexLibStatic ${WORKER_LIB} ${WORKER_DEP})
```
leogr, line 68: I've double-checked this in `debian:bullseye` (the same base image used by our release CI), rebuilding the plugin with and without this reorder. With `-lresolv` before `libworker.a`, GNU ld drops it (Debian's gcc passes `--as-needed` by default) and the resulting `.so` is identical to the shipped one: only `libc.so.6` as `NEEDED`, unversioned `UND __res_search`, and `dlopen` fails on glibc 2.31 with `undefined symbol: __res_search`. With the archive first, it gets `NEEDED: libresolv.so.2` plus `__res_search@GLIBC_2.2.5`, and it loads fine on both glibc 2.31 and 2.44 (the GLIBC floor stays at 2.28). So this reorder is indeed required.

Funny enough, this is also why my old attempt in #1036 didn't work 😅

plugins/container/cmake/modules/go-worker.cmake
```diff
@@ -25,6 +25,15 @@ if(APPLE)
...
     set(WORKER_DEP ${SECURITY_FRAMEWORK} ${RESOLV} ${CORE})
+elseif(CMAKE_HOST_SYSTEM_NAME STREQUAL "Linux")
+    # The same Go >= 1.20 cgo-resolver requirement applies on Linux: the net
+    # package references res_search from libresolv. On glibc >= 2.34 the
+    # symbol happens to resolve through libc's compat exports, which masks a
+    # missing dependency; on older glibc (Debian 11, RHEL/Rocky/Alma 8,
+    # Ubuntu 20.04) it lives only in libresolv.so.2, and without a DT_NEEDED
+    # entry the plugin fails to dlopen with:
+    #   undefined symbol: __res_search
+    set(WORKER_DEP resolv)
```
leogr, line 36: Just adding a data point on the "safe everywhere" side: `libresolv.so.2` is present in both our official images too (`falcosecurity/falco:0.44.1` is Wolfi with glibc 2.43, `0.44.1-debian` is Debian 12 with glibc 2.36), so the new `DT_NEEDED` is satisfied there as well :+1:

leogr (approved):
Thanks for this PR and for the thorough investigation in #1500! 🙏

LGTM. I double-checked the root cause (`libworker.a` pulls in Go's cgo resolver, and its `-lresolv` does not survive `-buildmode=c-archive`, so the consumer has to link it) and both halves of the fix, see inline comments.

A couple of notes, not blockers for this PR:
- None of our CI jobs load `libcontainer.so` on a glibc < 2.34 host (`build-linux` builds on bullseye but never `dlopen`s the result, and `falco-tests` runs inside `falcosecurity/falco:master-debian`, i.e. Debian 12 with glibc 2.36), so CI could not catch this. A quick guard in `build-linux` (e.g., `readelf -d libcontainer.so | grep -q libresolv.so.2`) would prevent regressions. Tracked in https://github.com/falcosecurity/plugins/issues/1512
- The website still documents the `LD_PRELOAD=/lib64/libresolv.so.2` workaround for RHEL 8 (https://falco.org/docs/setup/packages/#install-with-yum). Once we cut `v0.7.2` and bump the pin in Falco (tentatively for 0.45), we should update that note too. Tracked in https://github.com/falcosecurity/falco-website/issues/1582

cc @deepskyblue86

PR #1502, opened 2026-08-30
container: resubscribe to the CRI container events stream

### LINKED ISSUE #1499: container plugin: CRI event listener never re-subscribes after the runtime restarts
**Describe the bug**

The CRI engine subscribes once to `GetContainerEvents` in `Listen()` (`go-worker/pkg/container/cri.go`). When containerd restarts the stream ends, the producer goroutine closes `containerEventsCh`, the consumer returns and closes `outCh`, and `workerLoop` in `worker.go` just drops the case (`// Remove the stopped goroutine`). There is no retry, so from then on the plugin never learns about new containers through the `create`/`start` hooks.

The fallback in `fetcher.go` does not save it either: it retries `get()` for 150 ms total, and once it gives up on a container id it never asks again. So every container created after the runtime restart stays `<NA>` for its whole life, not just short-lived ones. On my cluster `stalwart-pg-1` has been running for a month and still logs `container.name=<NA>`.

**How to reproduce it**

1. Falco 0.44.1 with container plugin 0.7.1, `engines.cri` enabled on a containerd socket.
2. Restart containerd on the host (make sure the socket path is still reachable from Falco, for example by mounting its parent directory; with the current Helm chart the mount itself also breaks, see falcosecurity/charts#1052).
3. Start a new container and trigger any rule. `container.name`, `container.image.repository`, `k8s.ns.name`, `k8s.pod.name` are all `<NA>`. Containers created before the restart are still enriched.

**Expected behaviour**

`Listen()` should reconnect with a backoff when the event stream ends, and re-list running containers after reconnecting so nothing created during the gap is missed. It would also help if a failed fetch was not final, so a later lookup for the same container id gets another chance.

**Additional context**

Runtime restarts are normal on long-lived nodes (package upgrades with needrestart, containerd crashes, RKE2/k3s upgrades). Falco itself stays healthy and keeps emitting, so this failure is silent. The only signal I found is `falcosecurity_falcosidekick_falco_events_total{k8s_ns_name=""}` climbing for container-only rules.

Falco 0.44.1, container plugin 0.7.1, RKE2 v1.36.1 (containerd v2.2.3-k3s1), Helm chart 9.1.0. Same code on `main` as of today.

arpitjain099:
**What type of PR is this?**

/kind bug

/area plugins

`criEngine.Listen` subscribes to `GetContainerEvents` exactly once. When the runtime restarts, the stream ends, the producer goroutine closes `containerEventsCh`, the consumer returns and closes `outCh`, and `workerLoop` drops the case with a "Remove the stopped goroutine" comment. Nothing resubscribes, so from that point on the plugin never sees another create, start or delete event, and every container started after the restart stays `<NA>` for its whole life.

The producer now loops instead of returning: when `GetContainerEvents` comes back and the context is still alive, it logs and resubscribes after a backoff that starts at one second and doubles up to thirty. The events channel is only closed once the producer is really done, so the consumer keeps running across a restart.

Two smaller things fell out of that. Only the first subscription reports through `containerEventsErrorCh`, so the initialization check still returns the error a caller gets when events are disabled on the runtime, and that check now returns early only when the error is non-nil (before, a stream that ended cleanly inside the 10ms window made `Listen` return `nil, nil`). The consumer no longer needs its `containerEventsErrorCh` case, since that channel is no longer closed to signal the producer leaving.

This does not cover the second half of #1499, re-listing running containers after a reconnect so anything created during the gap is picked up. That needs a call on whether replaying containers the plugin may already know about is acceptable, and I would rather not guess at it here.

Fixes #1499

`TestCRIListenResubscribes` drives the real `Listen` with a runtime service that serves one event, ends the stream the way a dropped gRPC connection does, then serves a second event on the next subscription. It asserts both events arrive. On main it fails at the second one with "event channel closed instead of resubscribing after the stream ended". The backoff bounds are variables so the test can shorten them.

`go test -tags containers_image_openpgp -race ./pkg/container/ -run TestCRIListen` passes, existing subtests included. `TestContainerd` fails in my environment with or without this change, since it wants a live containerd socket.

```release-note
fix(plugins/container): resubscribe to the CRI container events stream when it ends, so containers created after a runtime restart are still enriched
```

### DIFF
--- a/plugins/container/go-worker/pkg/container/cri.go
+++ b/plugins/container/go-worker/pkg/container/cri.go
@@ -456,6 +456,17 @@ func (c *criEngine) List(ctx context.Context) ([]event.Event, error) {
 	return evts, nil
 }
 
+// resubscribeWait returns how long to wait before the next subscription
+// attempt. A subscription that outlived the ceiling was healthy, so the wait
+// starts over from the minimum instead of inheriting a backoff grown by
+// earlier restarts; otherwise the caller's current value stands.
+func resubscribeWait(current, lived time.Duration) time.Duration {
+	if lived > maxResubscribeBackoff {
+		return minResubscribeBackoff
+	}
+	return current
+}
+
 // Listen set up container created event loop by call to GetContainerEvents of the criEngine client
 // In case events have been disabled in the criEngine,
 // an error will be captured and passed to the caller.
@@ -469,6 +480,7 @@ func (c *criEngine) Listen(ctx context.Context, wg *sync.WaitGroup) (<-chan even
 		defer wg.Done()
 		backoff := minResubscribeBackoff
 		for first := true; ; first = false {
+			started := time.Now()
 			err := c.client.GetContainerEvents(ctx, containerEventsCh, nil)
 			if first {
 				containerEventsErrorCh <- err
@@ -476,10 +488,17 @@ func (c *criEngine) Listen(ctx context.Context, wg *sync.WaitGroup) (<-chan even
 			if ctx.Err() != nil {
 				return
 			}
+			backoff = resubscribeWait(backoff, time.Since(started))
 			// The stream ended without the context being cancelled, which is what
 			// a runtime restart looks like from here. Nothing resubscribes for us,
-			// so retry until the context goes away.
-			c.logger.LogAttrs(ctx, slog.LevelInfo, "container events stream ended, resubscribing",
+			// so retry until the context goes away. Only the first attempt is
+			// worth an operator's attention; a runtime that stays away would
+			// otherwise log every backoff period forever.
+			level := config.LevelTrace
+			if first {
+				level = slog.LevelInfo
+			}
+			c.logger.LogAttrs(ctx, level, "container events stream ended, resubscribing",
 				slog.String("socket", c.socket), slog.Duration("backoff", backoff), slog.Any("err", err))
 			select {
 			case <-ctx.Done():
--- a/plugins/container/go-worker/pkg/container/cri.go
+++ b/plugins/container/go-worker/pkg/container/cri.go
@@ -20,6 +20,13 @@ import (
 
 const maxCNILen = 4096
 
+// How long to wait before resubscribing to the container events stream, and the
+// ceiling the wait doubles up to. Variables so the tests can shorten them.
+var (
+	minResubscribeBackoff = 1 * time.Second
+	maxResubscribeBackoff = 30 * time.Second
+)
+
 func init() {
 	engineGenerators[typeCri] = newCriEngine
 }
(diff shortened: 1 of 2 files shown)


plugins/container/go-worker/pkg/container/cri.go
```diff
@@ -454,20 +461,42 @@ func (c *criEngine) List(ctx context.Context) ([]event.Event, error) {
...
-		containerEventsErrorCh <- c.client.GetContainerEvents(ctx, containerEventsCh, nil)
+		backoff := minResubscribeBackoff
+		for first := true; ; first = false {
+			err := c.client.GetContainerEvents(ctx, containerEventsCh, nil)
+			if first {
+				containerEventsErrorCh <- err
+			}
+			if ctx.Err() != nil {
+				return
+			}
```
leogr, line 490:
```suggestion
		for first := true; ; first = false {
			started := time.Now()
			err := c.client.GetContainerEvents(ctx, containerEventsCh, nil)
			if first {
				containerEventsErrorCh <- err
			}
			if ctx.Err() != nil {
				return
			}
			if time.Since(started) > maxResubscribeBackoff {
				// The previous subscription was healthy for a while, start over from the minimum.
				backoff = minResubscribeBackoff
			}
```

`backoff` only ever grows (see below), so after a handful of runtime restarts over the lifetime of the process every reconnect waits the full 30s, even if the previous stream had been healthy for days. Since containers started in that window only get enriched through the fetcher fallback, I'd reset the backoff once a subscription has lived long enough.

The threshold above is just an idea, wdyt? :thinking:

plugins/container/go-worker/pkg/container/cri.go
```diff
@@ -454,20 +461,42 @@ func (c *criEngine) List(ctx context.Context) ([]event.Event, error) {
...
+			// The stream ended without the context being cancelled, which is what
+			// a runtime restart looks like from here. Nothing resubscribes for us,
+			// so retry until the context goes away.
+			c.logger.LogAttrs(ctx, slog.LevelInfo, "container events stream ended, resubscribing",
+				slog.String("socket", c.socket), slog.Duration("backoff", backoff), slog.Any("err", err))
```
leogr, line 502: Nit. If the runtime never comes back, this line gets emitted every 30s indefinitely, and it is the only Info-level log in the go-worker (everything else is at trace).

Not a blocker, I'm just wondering if it should drop to Debug after the first attempt, or whether we prefer operators to notice. wdyt?

plugins/container/go-worker/pkg/container/cri_test.go
```diff
@@ -653,3 +654,108 @@ func TestCRIListen(t *testing.T) {
...
+	engine, err := newCriEngine(context.Background(), slog.Default(), endpoint)
+	require.NoError(t, err)
+	criEngine := engine.(*criEngine)
+
+	config.Load(fmt.Sprintf(`{"hooks": %d}`, config.HookCreate|config.HookStart|config.HookRemove))
+
+	// Keep the test from waiting out the production backoff.
+	minResubscribeBackoff, maxResubscribeBackoff = time.Millisecond, time.Millisecond
+	t.Cleanup(func() {
+		minResubscribeBackoff, maxResubscribeBackoff = time.Second, 30*time.Second
```
leogr:
```suggestion
	origMin, origMax := minResubscribeBackoff, maxResubscribeBackoff
	minResubscribeBackoff, maxResubscribeBackoff = time.Millisecond, time.Millisecond
	t.Cleanup(func() {
		minResubscribeBackoff, maxResubscribeBackoff = origMin, origMax
```

Just nit: restoring the saved values rather than the literals keeps the cleanup in sync if the defaults change.

leogr (commented):
Hey @arpitjain099, thanks for this fix (and welcome!) :pray:

The approach SGTM: the retry loop is bounded by the context, `containerEventsCh` is closed exactly once, and the new test really drives the second subscription. I ran it locally with `-race` and it passes, and on `main` the consumer would indeed close `outCh` after the first stream, as you described.

I also agree with leaving the re-list out of this PR. For the record, containers created while we are waiting to resubscribe are not entirely lost: the fetcher path (`AskForContainerInfo`) still picks them up on their first process event, as long as the runtime is reachable by then. So the real gap is the backoff window, which is one more reason for my comment below about resetting it.

Once that is addressed (see comments below), this is an easy LGTM from me.

N.B. The docker, podman, and containerd listeners do not resubscribe either. That is for a follow-up issue, not this PR :angel:

plugins/container/go-worker/pkg/container/cri.go
arpitjain099: Good catch, taken. I pulled the rule into a small helper so it is covered rather than implicit:

```go
func resubscribeWait(current, lived time.Duration) time.Duration {
	if lived > maxResubscribeBackoff {
		return minResubscribeBackoff
	}
	return current
}
```

and the loop now does `backoff = resubscribeWait(backoff, time.Since(started))` before waiting, so the reset applies to the reconnect that follows the healthy stream rather than the one after it. Your threshold is what I used: outliving the ceiling counts as healthy. `TestResubscribeWait` covers a stream that dies immediately keeping the grown value, one that outlives the ceiling starting over, and one just short of the ceiling not counting yet.

plugins/container/go-worker/pkg/container/cri.go
arpitjain099: Dropped to trace after the first attempt: `level := config.LevelTrace; if first { level = slog.LevelInfo }`. That keeps one Info line when the stream first drops, which is the part an operator wants to see, and a runtime that never comes back no longer writes an Info line every backoff period. It also matches the rest of the file, which is trace throughout.

plugins/container/go-worker/pkg/container/cri_test.go
arpitjain099: Done, it saves and restores the originals now.

arpitjain099:
All three addressed in f3ab472, and thanks for the note about `AskForContainerInfo` still picking up containers on their first process event. That does change how I read the gap: it is the enrichment delay inside the backoff window rather than a permanent loss, which is a better argument for the reset than the one I had.

Re-ran in a linux container with `-tags containers_image_openpgp -race`: `TestCRIListen`, `TestCRIListenResubscribes` and the new `TestResubscribeWait` all pass.

Happy to open the follow-up issue for the docker, podman and containerd listeners if you want it tracked.

leogr:
> Happy to open the follow-up issue for the docker, podman and containerd listeners if you want it tracked.

Yes please!

leogr (approved):
LGTM, thank you! :+1:

The red `libs-tests / test-e2e (system_deps)` job is the runner image regression tracked in #1510, unrelated to this change: same 16 assertion failures we see on every job that lands on the `20260819.586` image (also on PRs that do not touch this code), while the other two e2e variants passed on this very commit on the `20260828.587` image.

PR #1509, opened 2026-09-02
fix(.github): validate fork PR numbers in create-comment workflow

### LINKED ISSUE #1505: create-comment.yaml PR validation rejects all fork pull requests
The artifact validation added in #1226 aborts on every pull request opened from a fork, so the comment job fails for all outside contributions.

### What happens

```
Artifact PR number 1501 does not match any PR for commit d04fb7840b5b322af9012de15cb7d241b270111d. Aborting to prevent artifact poisoning.
```

The commit is correct and is still the head of #1501. It has not been force pushed.

### Why

The guard in `.github/workflows/create-comment.yaml` calls `listPullRequestsAssociatedWithCommit` against `context.repo`, which in a `workflow_run` event is always the base repository. For a fork PR the commit lives in the contributor's fork, so the base repo lookup returns nothing:

```
GET /repos/falcosecurity/plugins/commits/d04fb784.../pulls   -> []
GET /repos/ManuelFCastillo/plugins/commits/d04fb784.../pulls -> [#1501]
```

`valid_pr_numbers` is therefore always empty for fork PRs, so the check can never pass. It succeeds only for branches pushed directly to this repository.

### Possible fix

Resolve against the head repository from the `workflow_run` payload rather than `context.repo`:

```js
owner: context.payload.workflow_run.head_repository.owner.login,
repo: context.payload.workflow_run.head_repository.name,
```

I believe this keeps the property #1226 was protecting. The attack described there was a fork setting `NR` to an unrelated issue number, and that number still cannot be forged, because PR numbers are assigned by this repository and the lookup returns only PRs whose head is that commit. As belt and braces, the returned PR could also be checked to confirm its base repo is this one and that its head SHA matches.

Worth a review from whoever wrote the guard, since I may be missing a case it was defending against.

Seen on #1501. Affects any fork PR from 2026-03-02 onward. Happy to open a PR if the approach looks right.

ManuelFCastillo:
Fixes #1505

### Problem

The validation added in #1226 calls `listPullRequestsAssociatedWithCommit` against `context.repo`, which in a `workflow_run` event is always the base repository. For a pull request opened from a fork the head commit lives in the contributor's fork, so the base repo lookup returns an empty list, `valid_pr_numbers` is empty, and the check can never pass.

Every fork pull request has failed this job since 2026-03-02. Branches pushed directly to this repository are unaffected, which is why it went unnoticed.

### Change

Ports the approach already in use in falcosecurity/libs (falcosecurity/libs@0501075), as requested in #1505.

- For same-repo pull requests, `workflow_run.pull_requests` is populated and is used directly.
- For fork pull requests it is always empty, so fall back to `pulls.list` on `context.repo` filtered by `head: "<head_repository.owner.login>:<head_branch>"`.

Both values come from the `workflow_run` payload and are set by GitHub rather than by the fork.

### Why this preserves the guarantee from #1226

`pulls.list` is called on the base repository, so every candidate number is a pull request of this repository by construction. No additional `base.repo` or `head.sha` filtering is needed, which was the concern raised against querying the fork's `commits/{sha}/pulls` endpoint.

The failure message now names `owner:branch` rather than the commit SHA, matching the libs version.

### Verification

The block is byte-identical to the libs implementation. Confirmed the underlying behaviour against real pull requests:

```
GET /repos/falcosecurity/plugins/commits/<head_sha>/pulls
  #1501 (fork) -> []
  #1502 (fork) -> []
  #1489 (fork) -> []
  #1498 (same-repo) -> populated
```

### Note

`falcosecurity/rules` carries an identical workflow and needs the same change. I will open that separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/create-comment.yaml
+++ b/.github/workflows/create-comment.yaml
@@ -48,15 +48,30 @@ jobs:
 
             // Validate artifact PR number against the workflow_run context
             // to prevent artifact poisoning (cross-issue comment injection).
-            const head_sha = context.payload.workflow_run.head_sha;
-            const {data: associated_prs} = await github.rest.repos.listPullRequestsAssociatedWithCommit({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              commit_sha: head_sha,
-            });
-            const valid_pr_numbers = associated_prs.map(pr => pr.number);
+            //
+            // For same-repo PRs, workflow_run.pull_requests is populated.
+            // For fork PRs, pull_requests is always empty, so we fall back
+            // to searching for PRs by head repo owner and branch name.
+            const run_prs = context.payload.workflow_run.pull_requests;
+            let valid_pr_numbers = run_prs.map(pr => pr.number);
+
+            if (valid_pr_numbers.length === 0) {
+              // Fork PR: search using the head repo owner and branch from the
+              // workflow_run payload (these fields are set by GitHub, not the fork).
+              const head_owner = context.payload.workflow_run.head_repository.owner.login;
+              const head_branch = context.payload.workflow_run.head_branch;
+              const {data: matching_prs} = await github.rest.pulls.list({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                state: 'all',
+                head: `${head_owner}:${head_branch}`,
+                per_page: 10,
+              });
+              valid_pr_numbers = matching_prs.map(pr => pr.number);
+            }
+
             if (!valid_pr_numbers.includes(issue_number)) {
-              core.setFailed(`Artifact PR number ${issue_number} does not match any PR for commit ${head_sha}. Aborting to prevent artifact poisoning.`);
+              core.setFailed(`Artifact PR number ${issue_number} does not match any PR for head ${context.payload.workflow_run.head_repository.owner.login}:${context.payload.workflow_run.head_branch}. Aborting to prevent artifact poisoning.`);
               return;
             }
 


ManuelFCastillo:
While watching this one I noticed `create-comment.yaml` has drifted from libs in a second place, unrelated to the validation change here.

The `Download artifact` step filters for an artifact named `pr` and takes `[0]` unconditionally:

```js
var matchArtifact = artifacts.data.artifacts.filter((artifact) => {
  return artifact.name == "pr"
})[0];
// ...
artifact_id: matchArtifact.id,
```

When the triggering run produced no such artifact, `matchArtifact` is `undefined` and the step throws:

```
TypeError: Cannot read properties of undefined (reading 'id')
```

12 of the last 20 runs of this workflow have failed that way, all on `main`, going back to at least 2 September.

libs guards it (falcosecurity/libs@0501075):

```js
var matchArtifacts = artifacts.data.artifacts.filter((artifact) => {
  return artifact.name == "pr-perf"
});
if (!Array.isArray(matchArtifacts) || !matchArtifacts.length) {
  var process = require('process');
  process.exit();
}
```

It also guards the `unzip` and adds `&& !github.event.repository.fork` to the job condition.

Happy to fold the artifact guard into this PR, or leave it for a separate one if you would rather keep this change to the validation. Your call.

leogr:
Hey @ManuelFCastillo

Thanks for porting this! I've double-checked the block against https://github.com/falcosecurity/libs/commit/0501075c1fa79f23fb89dbf88140fbae8116a03e and it's identical, so LGTM :+1:

Good catch regarding the `Download artifact` step too. I confirm that `TypeError` is what we get every time `upload-pr-info` is skipped (i.e., when a PR doesn't touch any plugin), this PR included. However, I'd keep this one focused on the validation fix and port the artifact guard in a separate PR, so that each PR maps to the corresponding libs change. The same applies to `rules`. May you open them? 🙏

/lgtm
/approve

leogr (approved):

PR #1513, opened 2026-09-03
feat(ci): verify libcontainer.so links and loads on glibc < 2.34

### LINKED ISSUE #1512: container: CI should check that libcontainer.so links libresolv.so.2
**Motivation**

https://github.com/falcosecurity/plugins/issues/1500 went unnoticed because none of our CI jobs load `libcontainer.so` on a glibc < 2.34 host: `build-linux` builds on bullseye but never `dlopen`s the result, and `falco-tests` runs inside `falcosecurity/falco:master-debian` (Debian 12, glibc 2.36). So a missing `DT_NEEDED` on `libresolv.so.2` (fixed by https://github.com/falcosecurity/plugins/pull/1501) can silently come back with any Go or CMake change.

**Feature**

Add a guard to the container plugin CI (e.g. in `build-linux`), something like:

```sh
readelf -d libcontainer.so | grep -q 'NEEDED.*libresolv.so.2'
```

Even better, `dlopen` the built `.so` on a glibc < 2.34 image (e.g. `debian:bullseye`) as part of the tests.

**Additional context**

Needed for Falco 0.45, see https://github.com/falcosecurity/falco/issues/3967

/kind feature
/area plugins

ManuelFCastillo:
Fixes #1512

### Why CI missed #1500

No job both built against an old glibc and loaded the result:

- `build-linux` builds inside `debian:bullseye` (glibc 2.31) but never `dlopen`s what it produced.
- `falco-tests` does `dlopen`, but inside `falcosecurity/falco:master-debian` (Debian 12, glibc 2.36), where a missing `DT_NEEDED` on `libresolv.so.2` cannot fail, because those symbols moved into libc in glibc 2.34.

### Change

Both checks go in `build-linux`, which already runs on bullseye, so neither needs a new job, container or matrix entry, and both run for amd64 and arm64.

1. **`readelf -d`** asserts the `DT_NEEDED` entry is present. Fails fast and names the cause.
2. **`dlopen(..., RTLD_NOW)`** loads the built library on glibc 2.31. This is the check that would have caught #1500, and unlike the first it also catches a different symbol going missing on old glibc for an unrelated reason.

`RTLD_NOW` rather than `RTLD_LAZY` so every symbol is bound at load time; lazy binding would defer function resolution and could pass.

`readelf` and `gcc` both come from `build-essential`, already installed by the job.

### Verification

Built a probe library calling `res_search` two ways in `debian:bullseye` and ran both guards against each:

```
glibc: ldd (Debian GLIBC 2.31-13+deb11u14) 2.31

--- built with -lresolv ---
  readelf guard: PASS
  dlopen guard : PASS

--- built without -lresolv ---
  readelf guard: FAIL
  dlopen guard : FAIL
```

The second case is #1500 reproduced: a shared object links fine without `-lresolv` because undefined symbols are permitted at link time, then fails at `dlopen` on any host where `res_search` still lives in `libresolv`.

### Note

The issue also suggested loading on a glibc < 2.34 image as the stronger option. That is what step 2 does, and putting it in `build-linux` rather than a separate job means it runs against the artifact that job just built, rather than a rebuild.

/kind feature
/area plugins

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/container-ci.yaml
+++ b/.github/workflows/container-ci.yaml
@@ -4,10 +4,12 @@ on:
     branches: [main]
     paths:
       - "plugins/container/**"
+      - ".github/workflows/container-ci.yaml"
   push:
     branches: [main]
     paths:
       - "plugins/container/**"
+      - ".github/workflows/container-ci.yaml"
   workflow_dispatch:
 
 permissions:
@@ -114,7 +116,14 @@ jobs:
             return 0;
           }
           EOF
-          gcc -o /tmp/dlopen_check /tmp/dlopen_check.c -ldl
+          # The plugin pulls pthread_* and dl* from the Go runtime but declares
+          # no DT_NEEDED for them; under Falco they resolve from the global
+          # scope, because the Falco binary links both via libsinsp. Link the
+          # probe the same way so it fails on a genuinely missing dependency
+          # rather than on one Falco would have supplied. --no-as-needed is
+          # required because bullseye's gcc would otherwise drop libpthread,
+          # which the probe's own code never calls.
+          gcc -o /tmp/dlopen_check /tmp/dlopen_check.c -Wl,--no-as-needed -lpthread -ldl
           /tmp/dlopen_check ./libcontainer.so
           echo "loaded against $(ldd --version | head -n1)"
 
--- a/.github/workflows/container-ci.yaml
+++ b/.github/workflows/container-ci.yaml
@@ -84,6 +84,40 @@ jobs:
         working-directory: plugins/container
         run: make
 
+      # Issue #1500 shipped because no job both built against an old glibc and
+      # loaded the result: this one builds on bullseye but never dlopen'd it,
+      # and falco-tests dlopens inside Debian 12 (glibc 2.36), where a missing
+      # DT_NEEDED on libresolv cannot fail. This job is already on bullseye
+      # (glibc 2.31), so both checks cost nothing but the steps themselves.
+      - name: Check libcontainer.so links libresolv
+        working-directory: plugins/container
+        run: |
+          readelf -d libcontainer.so | grep -q 'NEEDED.*libresolv\.so\.2' || {
+            echo "libcontainer.so has no DT_NEEDED entry for libresolv.so.2."
+            echo "It will fail to load on hosts with glibc < 2.34. See #1500."
+            readelf -d libcontainer.so | grep NEEDED
+            exit 1
+          }
+
+      - name: Load libcontainer.so on glibc < 2.34
+        working-directory: plugins/container
+        run: |
+          cat > /tmp/dlopen_check.c <<'EOF'
+          #include <dlfcn.h>
+          #include <stdio.h>
+          int main(int argc, char **argv) {
+            void *handle = dlopen(argv[1], RTLD_NOW);
+            if (!handle) {
+              fprintf(stderr, "dlopen failed: %s\n", dlerror());
+              return 1;
+            }
+            return 0;
+          }
+          EOF
+          gcc -o /tmp/dlopen_check /tmp/dlopen_check.c -ldl
+          /tmp/dlopen_check ./libcontainer.so
+          echo "loaded against $(ldd --version | head -n1)"
+
       - name: Upload artifact
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:


.github/workflows/container-ci.yaml
```diff
@@ -84,6 +84,40 @@ jobs:
...
+          int main(int argc, char **argv) {
+            void *handle = dlopen(argv[1], RTLD_NOW);
+            if (!handle) {
+              fprintf(stderr, "dlopen failed: %s\n", dlerror());
+              return 1;
+            }
+            return 0;
+          }
+          EOF
+          gcc -o /tmp/dlopen_check /tmp/dlopen_check.c -ldl
```
leogr:
```suggestion
          gcc -o /tmp/dlopen_check /tmp/dlopen_check.c -Wl,--no-as-needed -lpthread -ldl
```

I ran this step against the real `libcontainer-amd64` artifact from https://github.com/falcosecurity/plugins/actions/runs/33775589712 inside `debian:bullseye` (glibc 2.31), and it fails:

```
dlopen failed: ./libcontainer.so: undefined symbol: pthread_mutex_trylock
```

The `.so` references `pthread_*` and `dl*` symbols (I believe they come from the Go runtime in `libworker.a`), but its only `DT_NEEDED` entries are `libresolv.so.2`, `libc.so.6`, and the loader. On glibc < 2.34 those symbols live in `libpthread.so.0` and `libdl.so.2`, so a bare probe cannot resolve them. It works when Falco loads the plugin since the Falco binary itself links `libpthread` and `libdl` (via `libsinsp`), and the plugin resolves them from the global scope. That is also why #1500 only surfaced `__res_search`: Falco does not link `libresolv`.

Linking the probe with `-Wl,--no-as-needed -lpthread -ldl` makes it load the plugin the way a Falco process does. The `--no-as-needed` part is needed since bullseye's `gcc` defaults to `--as-needed`, which drops `libpthread` from the probe (that is why a plain `-pthread` does not help). I verified that with this change the step passes on the same artifact, and it still fails on a `.so` calling `res_search` without `-lresolv`, so the #1500 case is still caught. `RTLD_NOW` can stay.

The proper fix would be linking `libpthread` and `libdl` explicitly in the plugin (e.g. `Threads::Threads` and `${CMAKE_DL_LIBS}` next to `resolv` in `go-worker.cmake`), so the `.so` is self-contained and the strict probe can stay as is. I did a quick simulation with a static archive and the resulting `.so` gets its own `DT_NEEDED` entries and loads with the strict probe, but I haven't rebuilt the actual plugin. However, that is a plugin change, so I'd keep it for a separate PR.

.github/workflows/container-ci.yaml
```diff
@@ -84,6 +84,40 @@ jobs:
         working-directory: plugins/container
         run: make
 
+      # Issue #1500 shipped because no job both built against an old glibc and
+      # loaded the result: this one builds on bullseye but never dlopen'd it,
+      # and falco-tests dlopens inside Debian 12 (glibc 2.36), where a missing
+      # DT_NEEDED on libresolv cannot fail. This job is already on bullseye
+      # (glibc 2.31), so both checks cost nothing but the steps themselves.
```
leogr, line 93: Nit (non-blocking). Since the workflow only triggers on `plugins/container/**`, the two new steps did not run on this PR (the green checks come from `ci.yaml` and CodeQL). May you add the workflow file itself to both `paths` lists (`pull_request` and `push`), so that changes like this one exercise themselves? :point_down:

```yaml
    paths:
      - "plugins/container/**"
      - ".github/workflows/container-ci.yaml"
```

That should also let the `dlopen` step run right here once the probe is fixed.

leogr (changes requested):
Hey @ManuelFCastillo

Thanks for this! Putting the checks in `build-linux` is the right call, since that job already runs on bullseye and covers arm64 too. The `readelf` guard is correct, and your analysis of why CI missed #1500 matches what I see: `falco-tests` runs in `falcosecurity/falco:master-debian`, which is Debian 12 with glibc 2.36.

However, I ran the two new steps against the real `libcontainer-amd64` artifact built from `main` (https://github.com/falcosecurity/plugins/actions/runs/33775589712) inside `debian:bullseye`. The `readelf` step passes, but the `dlopen` step fails with `undefined symbol: pthread_mutex_trylock`. As is, this would turn the next `plugins/container/**` PR red. Details and a minimal fix in the inline comments.

N.B. The workflow did not run on this PR, since it only triggers on `plugins/container/**` changes. That is why the checks here are green.

Thanks again :pray:

.github/workflows/container-ci.yaml
ManuelFCastillo: Applied, thanks. Good catch: I had only tested the guard against a .so built to be missing -lresolv, so I confirmed it caught the bug and never confirmed it passed a healthy build

Verified against the libcontainer-amd64 artifact from the run you linked, inside debian:bullseye: fails with undefined symbol: pthread_mutex_trylock as you described, loads cleanly with -Wl,--no-as-needed -lpthread -ldl, and still rejects a .so calling res_search without -lresolv

.github/workflows/container-ci.yaml
ManuelFCastillo: Added to both pull_request and push. build-linux ran here as a result, and both new steps passed on amd64 and arm64

leogr (approved):
LGTM, thanks for the quick turnaround! Both guards now pass against the real artifact on amd64 and arm64, and the workflow runs on its own changes too.

/approve

PR #1534, opened 2026-09-09
fix(plugins/container): bound the engine bootstrap and keep looking up unknown containers

### LINKED ISSUE #1269: bug(plugins/container): Fetcher retry timeout (150ms) causes permanent null metadata for containers under CRI load
**Describe the bug**

After upgrading from Falco 0.40.0 (libsinsp-based container engine) to 0.43.0 (container plugin), `k8s.ns.name`, `container.name`, `container.image.repository`, and other container metadata fields become permanently `null` for a significant number of
containers. On a single Kubernetes cluster (~3,300 active containers), null events increased from ~63/hour (0.40.0 baseline) to ~6,900/hour after the upgrade — a ~109x regression. The null fields persist indefinitely and do not self-heal.

The root cause is two-fold:

1. **The fetcher retry window is too short.** [`containerFetchRetryTimeout` is 150ms](https://github.com/falcosecurity/plugins/blob/f23b687eeb49fd2ec4d78e0bc6850d3d73973a25/plugins/container/go-worker/pkg/container/fetcher.go#L70-L71) with a fixed 30ms
retry interval (~5 retries). By comparison, libsinsp used exponential backoff (125ms–2000ms) over a ~20-second window. When the CRI runtime API takes longer than 150ms to respond (common under load), the fetcher silently drops the request.

2. **Failed lookups are never retried.** When the fetcher times out, `m_asked_containers` [records the container ID](https://github.com/falcosecurity/plugins/blob/f23b687eeb49fd2ec4d78e0bc6850d3d73973a25/plugins/container/src/plugin.cpp#L399) but no
metadata is ever delivered. Since [`on_new_process()` checks `m_asked_containers`](https://github.com/falcosecurity/plugins/blob/f23b687eeb49fd2ec4d78e0bc6850d3d73973a25/plugins/container/src/plugin.cpp#L388-L389) before issuing a new request, the
container is **never asked again** — resulting in permanently null metadata for the lifetime of that container.

**How to reproduce it**

1. Deploy Falco 0.43.0 with the container plugin (CRI engine enabled, containerd engine disabled) on a Kubernetes cluster with moderate-to-high pod churn
2. Observe Falco output fields for rules like `Launch Sensitive Mount Container` or `Write below root`
3. Under load, a significant portion of events will have `k8s.ns.name=null`, `container.name=""`, and other metadata fields empty
4. These null fields persist indefinitely — they do not self-heal over time

**Expected behaviour**

Container metadata should eventually be resolved for all containers, matching the behavior of libsinsp in Falco 0.40.0 where:
- Retries used exponential backoff over a ~20-second window
- Even failed lookups were stored in cache and could be retried on subsequent syscall events

**Environment**

- Falco version: 0.43.0
- Container plugin version: 0.6.3 (latest as of 2026-03-30)
- Container runtime: containerd (CRI-only mode, containerd engine disabled)
- Kubernetes: single cluster, ~3,300 active containers
- Observed null rate: ~6,900 events/hour on one cluster (vs ~63/hour on 0.40.0)

**Proposed fix**

We would like to contribute a fix addressing both issues:

1. **Replace fixed 30ms retry interval with exponential backoff** in the [fetcher](https://github.com/falcosecurity/plugins/blob/f23b687eeb49fd2ec4d78e0bc6850d3d73973a25/plugins/container/go-worker/pkg/container/fetcher.go#L69-L107). This requires
changing the channel message type from `chan string` to a struct carrying retry count and backoff state. Target: 5 retries with 125ms/250ms/500ms/1000ms/2000ms intervals (~4 seconds total), matching libsinsp's strategy.

2. **Add TTL-based expiry to `m_asked_containers`** in [plugin.h](https://github.com/falcosecurity/plugins/blob/f23b687eeb49fd2ec4d78e0bc6850d3d73973a25/plugins/container/src/plugin.h#L130) /
[plugin.cpp](https://github.com/falcosecurity/plugins/blob/f23b687eeb49fd2ec4d78e0bc6850d3d73973a25/plugins/container/src/plugin.cpp#L388-L399). Change from `std::unordered_set<std::string>` to `std::unordered_map<std::string, timestamp>` so that entries
  expire after a configurable TTL, allowing re-lookup for containers whose initial fetch failed.

**Additional context**

Comparison of retry behavior:

| | libsinsp (0.40.0) | container plugin (current) |
|---|---|---|
| Retry interval | Exponential: 125, 250, 500, 1000, 2000ms | Fixed 30ms |
| Total retry window | ~20 seconds | 150ms |
| On timeout | Stores FAILED result in cache | Silent drop, nothing cached |
| Re-lookup | Within retry window (sufficient time) | `m_asked_containers` blocks permanently |

Additionally, the retry goroutine in the fetcher uses a [blocking channel send](https://github.com/falcosecurity/plugins/blob/f23b687eeb49fd2ec4d78e0bc6850d3d73973a25/plugins/container/go-worker/pkg/container/fetcher.go#L107) for re-enqueue, while
[`AskForContainerInfo()` uses non-blocking send](https://github.com/falcosecurity/plugins/blob/f23b687eeb49fd2ec4d78e0bc6850d3d73973a25/plugins/container/go-worker/worker_api.go#L128-L131). Simply increasing the timeout without changing to exponential
backoff would risk goroutine accumulation and channel saturation.

Related issue: #1014

### LINKED ISSUE #1487: bug(plugins/container): unbounded blocking in engine bootstrap hangs Falco startup
**Describe the bug**

`StartWorker` bootstraps all enabled container engines **synchronously, on the caller's thread**, before returning. The context it uses is derived from `context.Background()` with cancel only, so it has **no deadline**. As a consequence, a runtime socket that accepts a connection but never answers blocks the call forever.

Since `StartWorker` is invoked from `start_async_events`, which libs calls from `sinsp::open_common()` on Falco's **main thread**, this hangs Falco's startup permanently, before the event loop is even reached.

Reported downstream at https://github.com/falcosecurity/falco/issues/3953

The chain:

- `StartWorker` runs `generator(ctx)` and then `engine.List(ctx)` for each engine, serially and inline, with no deadline 👇
  https://github.com/falcosecurity/plugins/blob/21d3ee6df859604a93f3ff41e9b9f0b56495740e/plugins/container/go-worker/worker_api.go#L66-L83
- `newPodmanEngine` calls `bindings.NewConnection`, which pings the service with a `GET /_ping`. For `unix://` URIs, the podman bindings build an `http.Client` with **no `Timeout`** and a `net.Dialer` with no timeout 👇
  https://github.com/falcosecurity/plugins/blob/21d3ee6df859604a93f3ff41e9b9f0b56495740e/plugins/container/go-worker/pkg/container/podman.go#L33-L39
- `podmanEngine.List` **discards its `ctx` parameter** and uses the connection context instead 👇
  https://github.com/falcosecurity/plugins/blob/21d3ee6df859604a93f3ff41e9b9f0b56495740e/plugins/container/go-worker/pkg/container/podman.go#L188
- Sockets are not reachability-probed. `Generators()` creates an engine for any path whose `os.Stat` does not return `NotExist` 👇
  https://github.com/falcosecurity/plugins/blob/21d3ee6df859604a93f3ff41e9b9f0b56495740e/plugins/container/go-worker/pkg/container/engine.go#L61-L85

Affected engines:

| Engine | Connect timeout | `List` bounded |
|---|---|---|
| `podman` | **none** | **no** (ignores `ctx`) |
| `docker` | **none** | **no** |
| `containerd` | **none** | **no** |
| `cri` | 5s | yes |

So `cri` is the only one already protected 👇
https://github.com/falcosecurity/plugins/blob/21d3ee6df859604a93f3ff41e9b9f0b56495740e/plugins/container/go-worker/pkg/container/cri.go#L43

**Aggravating factor: rootless podman auto-discovery**

When `engines.podman.sockets` is empty, the plugin adds `/run/podman/podman.sock` **and scans `/run/user/*/podman/podman.sock`**, enabling everything it finds 👇
https://github.com/falcosecurity/plugins/blob/21d3ee6df859604a93f3ff41e9b9f0b56495740e/plugins/container/src/plugin_config.cpp#L66-L93

That scan is active on host installs, since Falco ships `engines` unset ("We use default config values for `engines` key" in `falco.yaml`).

Those sockets are socket-activated by `systemd`, per user. Falco runs as root, so `connect()` succeeds because `systemd` owns the listener, but the response only arrives once that user's `podman.service` starts. If the user manager is not running, no response ever arrives.

N.B. The Helm chart pins `podman.sockets: ["/run/podman/podman.sock"]`, which suppresses the scan. So Kubernetes deployments are largely shielded, and **host RPM/DEB installs are the exposed configuration**.

**How to reproduce it**

1. Install Falco from the RPM/DEB packages, so that `engines` is left at its defaults
2. Make a unix socket available at `/run/user/<uid>/podman/podman.sock` that accepts connections but never writes a response. A socket-activated rootless `podman.service` that cannot start reproduces this naturally (this is what the reporter hit). For a synthetic repro, something like `socat UNIX-LISTEN:<path>,fork SYSTEM:'sleep infinity'` should do, though I have not verified that exact invocation yet
3. Start Falco

Falco logs up to `Trying to open the right engine!` and then hangs indefinitely. `SIGINT` and `SIGTERM` have no effect, since they only set a flag that the event loop polls, so only `SIGKILL` works.

**Expected behaviour**

An unresponsive or misbehaving runtime socket must never be able to block Falco's startup. The engine bootstrap should be bounded, and an engine that does not answer in time should be reported as unavailable instead of hanging.

**Environment**

- Container plugin version: **0.7.1**, and also reproducible on `main` (no timeout changes have landed on `go-worker` since 0.7.1)
- Falco version: 0.44.1
- libs version: 0.25.4
- OS: Rocky Linux 10.2, kernel `6.12.0-211.42.1.el10_2.x86_64`
- Installation method: RPM, also reported with the Docker/Podman images
- Drivers: reproduced with both `kmod` and `modern_ebpf`, since the plugin bootstrap runs before the driver matters

**Proposed fix**

In order of risk, from lowest:

1. **Bound the bootstrap.** Wrap each `generator(ctx)` and `engine.List(ctx)` in a `context.WithTimeout` inside `StartWorker`. This just normalizes the behaviour to what `cri` already does, and the timeout should be configurable.
2. **Honour the passed `ctx`.** `podmanEngine.List` currently discards it, so (1) alone would have no effect for podman. The other engines should be audited for the same problem.
3. **Set explicit client timeouts** where the SDK allows it, i.e. `client.WithTimeout` for docker and `containerd.WithTimeout` for containerd. The podman bindings own their `http.Client`, so the ctx deadline from (1) is the only lever there.
4. **Follow-up, larger change: do not block startup at all.** Move the pre-existing container enumeration into the worker goroutine and let `StartWorker` return immediately. N.B. this needs care, since `capture_open` enriches the initial thread table right afterwards, so it would race against the merge of `s_preexisting_containers`. I'd do 1-3 first and treat this one separately.

**Additional context**

There is also a design question worth discussing. Auto-enabling every `/run/user/*/podman/podman.sock` found on disk gives a root-running Falco a hard dependency on arbitrary users' socket-activated services. At the very least, those sockets should be reachability-probed with a short timeout before being enabled. Arguably, the scan should be opt-in. wdyt? :thinking:

cc @deepskyblue86 @FedeDP

leogr:
**What type of PR is this?**

/kind bug

/area plugins

Two fixes for the container plugin, cumulated here so they ship together in `0.7.3` 👇

**1. Bound the engine bootstrap (#1487)**

`StartWorker` bootstraps every enabled engine synchronously with a context that has no deadline, and the podman client has no timeout either: a runtime socket that accepts the connection but never answers blocks the call forever, and since libs calls it from `sinsp::open_common()` on the main thread, Falco hangs at startup and ignores SIGTERM.

- new `engine_timeout` init config key (seconds, default `10`, `0` disables it), documented in the README
- each engine's initial `List` runs under the timeout; an engine that does not answer is logged and skipped
- an engine that answers but does not finish in time is kept: the containers inspected so far are loaded, the others (and, for containerd, the namespaces not enumerated yet) are looked up in the background right after startup, one at a time and behind the lookups the events ask for, so no partial metadata ever enters the cache
- the podman `/_ping` is bounded, and podman `List`/`get` honour the caller's context

**2. Keep looking up a container the runtime does not know yet (#1269)**

The go-worker retried a missed lookup every 30 ms for 150 ms and then dropped it, and the plugin never asked again: a container the runtime did not know yet (e.g. a loaded CRI) stayed without metadata for the whole plugin lifetime.

- the fetcher retries with an exponential backoff (125 ms to 2 s, about 4 s in total), with a single goroutine and a single timer, deduplicating pending ids and capping them at 1024
- the plugin's "already asked" set expires after 10 s, so an unknown container is asked again on its next process event; entries are also dropped on `container_removed`
- two latent shutdown races in the old fetcher are gone as a side effect

Hot path: `on_new_process` stays O(1), and goroutines, timers and memory are bounded. Unit tests cover the black-hole socket, the cut listing for every engine, the background recovery (including native containerd with full ids across namespaces), the retry schedule, dedup and a 5000-miss stress test; benchmarks live in `fetcher_test.go`. Verified end to end with Falco in `nodriver` mode against a socket that accepts and never answers.

N.B. `engine_timeout` bounds the whole initial listing per engine: on hosts with a huge number of containers, the ones not inspected in time get their metadata a little later, as the background lookups reach them; raise it (or set `0`) to have them all at startup. With podman the timeout can be exceeded by up to about 0.6 s, since its client retries a failed request three times.

Fixes #1487
Fixes #1269
Ref falcosecurity/falco#3953

The PR went through an independent correctness and performance review and four more review rounds on the startup path; each round's fix is its own commit. The suggestions that change runtime behaviour beyond these two fixes (global vs per-engine timeout, retrying an engine that timed out on `List`, a growing TTL for ids that never resolve, a deadline on steady-state lookups) are tracked in #1535, so that `0.7.3` ships the two fixes as reviewed.

- the backoff schedule, the pending cap and the 10 s TTL are constants on purpose (a knob can come later if a real use case shows up); the TTL must stay above the go-worker retry window
- containers that never resolve are re-asked every 10 s (up to 6 lookups per engine per ask) instead of once ever: bounded, miss path only
- native containerd lookups are exact-id, so the containers left over by a cut listing keep their full id, namespace and engine for the background lookups

```release-note
fix(plugins/container): bound the startup connection to each container engine with the new `engine_timeout` init config (default 10s, 0 disables it); an engine that does not answer in time is skipped with a warning instead of hanging Falco startup
fix(plugins/container): keep looking up a container the runtime does not know yet (exponential backoff up to about 4s, then ask again after 10s) instead of leaving it without metadata forever
```

### DIFF
--- a/plugins/container/README.md
+++ b/plugins/container/README.md
@@ -137,6 +137,7 @@ plugins:
       label_max_len: 100 # (optional, default: 100; container labels larger than this won't be reported)
       with_size: false # (optional, default: false; whether to enable container size inspection, which is inherently slow)
       hooks: ['create', 'start'] # (optional, default: 'create'. Some fields might not be available in create hook, but we are guaranteed that it gets triggered before first process gets started)
+      engine_timeout: 10 # (optional, default: 10; seconds to wait for a container engine to answer at startup, 0 disables the timeout)
       engines:
         docker:
           enabled: true
@@ -159,6 +160,8 @@ plugins:
 
 load_plugins: [container]
 ```
+
+At startup, the plugin connects to each enabled engine socket and lists the pre-existing containers. An engine that does not answer within `engine_timeout` seconds (default: 10) is skipped with a warning, so that an unresponsive runtime socket (for example a socket-activated service whose backend is gone) cannot block Falco's startup. Setting `engine_timeout` to `0` disables the timeout: the plugin then waits for each engine indefinitely.
 
 ### Rules
 
--- a/plugins/container/README.md
+++ b/plugins/container/README.md
@@ -161,7 +161,7 @@ plugins:
 load_plugins: [container]
 ```
 
-At startup, the plugin connects to each enabled engine socket and lists the pre-existing containers. An engine that does not answer within `engine_timeout` seconds (default: 10) is skipped with a warning, so that an unresponsive runtime socket (for example a socket-activated service whose backend is gone) cannot block Falco's startup. An engine that answers but does not finish listing its containers in time is kept: the containers inspected so far are loaded, and the others are looked up on their first event, like any container the plugin does not know yet, rather than being loaded with partial metadata. Setting `engine_timeout` to `0` disables the timeout: the plugin then waits for each engine indefinitely.
+At startup, the plugin connects to each enabled engine socket and lists the pre-existing containers. An engine that does not answer within `engine_timeout` seconds (default: 10) is skipped with a warning, so that an unresponsive runtime socket (for example a socket-activated service whose backend is gone) cannot block Falco's startup. An engine that answers but does not finish listing its containers in time is kept: the containers inspected so far are loaded, and the others are looked up in the background right after startup, one at a time and behind the lookups the events ask for, rather than being loaded with partial metadata; their events carry no container metadata until then. Setting `engine_timeout` to `0` disables the timeout: the plugin then waits for each engine indefinitely.
 
 N.B. With podman the timeout can be exceeded by up to about 0.6 seconds, since the podman client retries a failed request three times with fixed pauses in between.
 
--- a/plugins/container/README.md
+++ b/plugins/container/README.md
@@ -161,7 +161,9 @@ plugins:
 load_plugins: [container]
 ```
 
-At startup, the plugin connects to each enabled engine socket and lists the pre-existing containers. An engine that does not answer within `engine_timeout` seconds (default: 10) is skipped with a warning, so that an unresponsive runtime socket (for example a socket-activated service whose backend is gone) cannot block Falco's startup. Setting `engine_timeout` to `0` disables the timeout: the plugin then waits for each engine indefinitely.
+At startup, the plugin connects to each enabled engine socket and lists the pre-existing containers. An engine that does not answer within `engine_timeout` seconds (default: 10) is skipped with a warning, so that an unresponsive runtime socket (for example a socket-activated service whose backend is gone) cannot block Falco's startup. An engine that answers but does not finish listing its containers in time is kept: the containers inspected so far are loaded, and the others are looked up on their first event, like any container the plugin does not know yet, rather than being loaded with partial metadata. Setting `engine_timeout` to `0` disables the timeout: the plugin then waits for each engine indefinitely.
+
+N.B. With podman the timeout can be exceeded by up to about 0.6 seconds, since the podman client retries a failed request three times with fixed pauses in between.
 
 ### Rules
 
--- a/plugins/container/README.md
+++ b/plugins/container/README.md
@@ -161,7 +161,9 @@ plugins:
 load_plugins: [container]
 ```
 
-At startup, the plugin connects to each enabled engine socket and lists the pre-existing containers. An engine that does not answer within `engine_timeout` seconds (default: 10) is skipped with a warning, so that an unresponsive runtime socket (for example a socket-activated service whose backend is gone) cannot block Falco's startup. An engine that answers but does not finish listing its containers in time is kept: the containers inspected so far are loaded, and the others are looked up in the background right after startup, one at a time and behind the lookups the events ask for, rather than being loaded with partial metadata; their events carry no container metadata until then. Setting `engine_timeout` to `0` disables the timeout: the plugin then waits for each engine indefinitely.
+At startup, the plugin connects to each enabled engine socket and lists the pre-existing containers. An engine that does not answer within `engine_timeout` seconds (default: 10) is skipped with a warning, so that an unresponsive runtime socket (for example a socket-activated service whose backend is gone) cannot block Falco's startup. An engine that answers but does not finish listing its containers in time is kept: the containers inspected so far are loaded, and the others are looked up in the background right after startup, one at a time and behind the lookups the events ask for; their events carry no container metadata until then.
+
+For containerd, this also includes namespaces whose container listing was interrupted or had not started. Their listing is retried in the background until it succeeds, without requiring a new process event. Each attempt uses `engine_timeout` and stops when capture stops. Setting `engine_timeout` to `0` disables the timeout: startup and background namespace listings then wait indefinitely unless cancelled.
 
 N.B. With podman the timeout can be exceeded by up to about 0.6 seconds, since the podman client retries a failed request three times with fixed pauses in between.
 
--- a/plugins/container/go-worker/bootstrap_test.go
+++ b/plugins/container/go-worker/bootstrap_test.go
@@ -25,7 +25,8 @@ type listEngine struct {
 	unresponsive bool
 	// notInspected, when set, makes List wait for the context to be done and
 	// return the events with a ListIncompleteError for these containers.
-	notInspected []container.ContainerRef
+	notInspected  []container.ContainerRef
+	notEnumerated []string
 }
 
 func (e *listEngine) Name() string { return e.name }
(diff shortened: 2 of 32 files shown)


leogr:
Hey @deepskyblue86

This one is now ready for review 👉 we'd like to ship these two fixes in `0.7.3` for Falco 0.45.0, so any chance you can take a look in the next day or so? 🙏

leogr:
/hold

I may have found some issues. Fixing.

deepskyblue86 (approved):
Remarkable.
/approve

Issue #1535, opened 2026-09-09
container: follow-ups from the 0.7.3 engine bootstrap and lookup retry review


leogr:
**Describe the enhancement**

Follow-ups from the independent review of https://github.com/falcosecurity/plugins/pull/1534 (container plugin `0.7.3`). None of them blocks the release, they change runtime behavior, so they are better handled on their own 👇

- **Per-engine bootstrap timeout**: `engine_timeout` bounds each engine's initial listing separately, so the worst-case startup with N unresponsive engines is about 2 × N × `engine_timeout`. Decide whether the bound should be global instead, or document the current behavior more prominently.
- **Engine skipped on a `List` timeout is dropped for the lifetime of the plugin**: an engine that answered `/_ping` but timed out on the initial `List` is never retried, while a partial listing keeps the engine. Consider retrying the engine later, or closing it explicitly and logging that it is gone for good.
- **Bootstrap cancelled during shutdown is reported as a timeout**: the skip log uses the listing context error, so a shutdown while an engine is being bootstrapped logs as "did not answer within the engine timeout". Distinguish `ctx.Err()` from the listing context error.
- **Unresolvable container ids are re-asked every 10 s forever**: a container id the runtime never knows (for example a sandbox container) costs one lookup per engine every 10 s on the miss path. Consider a growing TTL (10 s, 20 s, 40 s, capped) or a maximum number of rounds.
- **Steady-state lookups have no deadline**: `lookup` uses a background context, so a socket that hangs after startup can block `StopWorker` and the fetcher goroutine. Bound each lookup with a timeout derived from `engine_timeout`.
- **Retry window and re-ask period are constants**: the 125 ms to 2 s backoff (about 4 s total) and the 10 s `asked_containers` TTL are constants on purpose in `0.7.3`. Make them configurable only if a real use case shows up (https://github.com/falcosecurity/plugins/issues/1269 asked for it).

**Additional context**

The independent review also noted that `erase()` in `asked_containers` leaves `m_order` untouched until the lazy purge reaches the entry, which is documented in the header now, and that the debug log volume on the miss path may be worth a look.

Ref https://github.com/falcosecurity/plugins/pull/1534
Ref https://github.com/falcosecurity/plugins/issues/1487
Ref https://github.com/falcosecurity/plugins/issues/1269

/kind feature
/area plugins

kokhlo:
I can take this one.

[...]

PR #1538, opened 2026-09-11
fix(plugins/container): bound each steady-state lookup with the engine timeout

kokhlo:
## Summary

The fetcher's `lookup` passed its background context to every engine probe, so a runtime socket that accepts connections but never answers **after startup** blocked the single serving goroutine for good: live requests, retries and the deferred startup lookups all stalled, and `StopWorker` could not reap the goroutine.

This is the **"Steady-state lookups have no deadline"** follow-up of #1535. The fix bounds each engine probe with the configured `engine_timeout`, the same `WithEngineTimeout` the bootstrap listing (`bootstrapEngines`) and the deferred namespace enumeration (`enumerateNamespace`) already use, so a hanging engine only costs its own timeout and the next engine is consulted with a fresh bound. A zero timeout keeps the previous unbounded behavior, as elsewhere.

Closes the steady-state lookup item of #1535.

## Testing

Regression test (`TestFetcherRequestLookupHonoursEngineTimeout`, `TestFetcherDeferredLookupHonoursEngineTimeout`): a hanging engine never answers; the lookup must give way at the engine timeout and the next engine must serve the container. On main the first test wedges forever (fetcher goroutine stuck in `lookup`, cleanup blocked in `wg.Wait()`); with the fix:

```
=== RUN   TestFetcherRequestLookupHonoursEngineTimeout
--- PASS: TestFetcherRequestLookupHonoursEngineTimeout (1.00s)
=== RUN   TestFetcherDeferredLookupHonoursEngineTimeout
--- PASS: TestFetcherDeferredLookupHonoursEngineTimeout (1.00s)
ok    github.com/falcosecurity/plugins/plugins/container/go-worker/pkg/container    2.015s
```

Focused regression run with upstream's flags (Linux, `-tags containers_image_openpgp -race`):

```
go test -tags containers_image_openpgp -race -count=1 -run 'TestFetcher|TestBootstrap|TestDockerList|TestContainerdList' ./pkg/container/
ok    github.com/falcosecurity/plugins/plugins/container/go-worker/pkg/container    6.168s
```

Note: the full `make test` suite includes `TestContainerd`, which needs a live containerd socket; it fails identically on clean main in the rootless container used for local verification, so it is environmental, not a regression of this change. The new tests and all fetcher/bootstrap/timeout tests pass with `-race`.

### DIFF
--- a/plugins/container/go-worker/pkg/container/fetcher.go
+++ b/plugins/container/go-worker/pkg/container/fetcher.go
@@ -395,14 +395,20 @@ func (f *fetcher) forgetFailedDeferred(id string) {
 
 // lookup uses the known runtime identity of a deferred container, or probes
 // the engines for an unknown ID. Success publishes the event and releases
-// any deferred identity.
+// any deferred identity. Each engine answers within the engine timeout, as
+// at bootstrap, so a socket that hangs after startup cannot block the
+// serving goroutine for good.
 func (f *fetcher) lookup(ctx context.Context, containerId string, outCh chan<- event.Event) bool {
 	var evt *event.Event
 	if d, ok := f.deferredLookups[containerId]; ok {
-		evt, _ = d.get(f.ctx)
+		lctx, cancel := WithEngineTimeout(ctx)
+		evt, _ = d.get(lctx)
+		cancel()
 	} else {
 		for _, e := range f.getters {
-			evt, _ = e.get(f.ctx, containerId)
+			lctx, cancel := WithEngineTimeout(ctx)
+			evt, _ = e.get(lctx, containerId)
+			cancel()
 			if evt != nil {
 				break
 			}
--- a/plugins/container/go-worker/pkg/container/fetcher_lookup_timeout_test.go
+++ b/plugins/container/go-worker/pkg/container/fetcher_lookup_timeout_test.go
@@ -1,0 +1,97 @@
+package container
+
+import (
+	"context"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/falcosecurity/plugins/plugins/container/go-worker/pkg/event"
+)
+
+// hangingGetter accepts a lookup and never answers it until its context is
+// done, like a runtime socket that hangs after startup.
+type hangingGetter struct {
+	mu      sync.Mutex
+	lookups int
+}
+
+func (g *hangingGetter) get(ctx context.Context, _ string) (*event.Event, error) {
+	g.mu.Lock()
+	g.lookups++
+	g.mu.Unlock()
+	<-ctx.Done()
+	return nil, ctx.Err()
+}
+
+func (g *hangingGetter) count() int {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	return g.lookups
+}
+
+// newTestFetcherGetters is newTestFetcher with an explicit getter list.
+func newTestFetcherGetters(getters []getter, fetchCh chan string, backoff []time.Duration, maxPending int, deferred []string) *fetcher {
+	f := &fetcher{
+		getters:         getters,
+		ctx:             context.Background(),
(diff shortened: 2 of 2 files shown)


plugins/container/go-worker/pkg/container/fetcher.go
```diff
@@ -395,14 +395,20 @@ func (f *fetcher) forgetFailedDeferred(id string) {
...
 	if d, ok := f.deferredLookups[containerId]; ok {
-		evt, _ = d.get(f.ctx)
+		lctx, cancel := WithEngineTimeout(ctx)
+		evt, _ = d.get(lctx)
+		cancel()
 	} else {
 		for _, e := range f.getters {
-			evt, _ = e.get(f.ctx, containerId)
+			lctx, cancel := WithEngineTimeout(ctx)
+			evt, _ = e.get(lctx, containerId)
```
c2ndev, line 410: Can we check `lctx.Err()` before `cancel()` in both paths and treat an expired lookup as failed? Docker's getter can return a non-nil event after `ImageInspect` times out, so this currently publishes incomplete metadata and skips retries. The C++ cache then considers the container resolved.

Reproduced with `engine_timeout=1` and a 1.5s image response: the baseline returns complete metadata; this PR publishes empty image ID/digest with no retry, for both live and deferred lookups.

Please keep the existing handling of ordinary best-effort errors, and add coverage for a timeout during enrichment followed by a successful retry, including preservation of the deferred identity.

plugins/container/go-worker/pkg/container/fetcher_lookup_timeout_test.go
```diff
@@ -0,0 +1,97 @@
...
+// to a hanging engine, then asks for a live container: the deferred lookup
+// must return once the engine timeout expires so the request can be served,
+// instead of stalling the single fetcher goroutine.
+func TestFetcherDeferredLookupHonoursEngineTimeout(t *testing.T) {
+	setEngineTimeout(t, 1)
+	hanging := &hangingGetter{}
+	healthy := newCountingGetter(1)
+	fetchCh := make(chan string)
+	f := newTestFetcherGetters([]getter{hanging, healthy}, fetchCh, fastBackoff, maxPendingFetches, []string{"aabbccddeeff"})
+	outCh := runFetcher(t, f)
```
c2ndev, line 85: Can we wait for the deferred lookup to enter the getter before sending `late`? The live request can run first, and `hanging.count()` counts entries, not completed calls. An unbounded deferred lookup can therefore pass this assertion and only unblock during cleanup.

Replacing just the deferred `WithEngineTimeout(ctx)` with `context.WithCancel(ctx)` still passes this test 5/5 times with `GOMAXPROCS=1` and `-race`. A signal keyed to the deferred ID would make the ordering explicit and let the test verify progress before cleanup cancellation.

c2ndev (commented):
Hi @kokhlo, thank you for your contribution!
I found two issues while testing this: a timeout during metadata enrichment can leave incomplete data cached without retries, and the deferred lookup test passes even when its timeout is removed. Added inline comments with repro details and suggested fixes.
Let me know if something is not clear :)

plugins/container/go-worker/pkg/container/fetcher.go
```diff
@@ -395,14 +395,20 @@ func (f *fetcher) forgetFailedDeferred(id string) {
 
 // lookup uses the known runtime identity of a deferred container, or probes
 // the engines for an unknown ID. Success publishes the event and releases
-// any deferred identity.
+// any deferred identity. Each engine answers within the engine timeout, as
```
leogr: May you update the `engine_timeout` descriptions in the init schema and README? They currently describe startup (and background namespace listing), while this change also applies the timeout to live and deferred container lookups.

leogr (commented):
The per-engine timeout looks like the right direction. I agree with the two points already raised by @c2ndev. One small documentation note inline.

